PR #291, opened 2013-03-08
Fix to support openLDAP / posix style LDAP

memotype:
Config options added and user authentication code updated to support a posix
/ OpenLDAP type of query.

TODO: Welcome message shows "uid=username" instead of just username, or
firstname.


abcbarryn:
So does it still say "Welcome uid=foo" instead of "Welcome foo" or was that addressed also?

PR #1005, opened 2015-07-05
fix #950

### LINKED ISSUE #950: can not create Admin account
I can't install teampass, connection to the database is 0k, user is 0k, I try with root user, and tpadmin, and it's the same error. All privileges are granted, and I can do everything from console, but from GUI does not work.
teampassadminaccount

dorianmuthig:
Notes:
* there are no default avatars available, otherwise it might be better to set those instead, however, checking if the value is NULL and then selecting the default one seems a much more sane approach

* I might have messed up indentation, I use spaces instead of tabs (with one tab being 4 spaces) and the file was inconsistent in what was used (spaces or tabs)
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/nilsteampassnet/TeamPass/pull/1005?approve=1'></a>&nbsp;<a href='https://github.com/nilsteampassnet/TeamPass/pull/1005'></a>
<a href='#crh-end'></a>

### DIFF
--- a/install/install.queries.php
+++ b/install/install.queries.php
@@ -20,15 +20,15 @@ $_SESSION['db_encoding'] = "utf8";
 $_SESSION['CPM'] = 1;
 
 function chmod_r($dir, $dirPermissions, $filePermissions) {
-	$dp = opendir($dir);
-	$res = true;
-	while($file = readdir($dp)) {
-		if (($file == ".") || ($file == ".."))
-			continue;
+    $dp = opendir($dir);
+    $res = true;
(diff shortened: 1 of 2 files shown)


nilsteampassnet:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/nilsteampassnet/TeamPass/pull/1005'>Refresh</a>.<a href='#crh-update-none'></a>

nilsteampassnet:
Thanks

PR #1015, opened 2015-07-24
Added randomization of OTV IDs

alikkon:
A small snippet was added to sources/items.queries.php from lines 2713 to 2728. The new code creates a randomized OTP identifier, instead of using the sequential one that was previously in use. If the generated random identifier already exists, we keep creating new randoms until we find one that doesn't. While this can theoretically result in a link working twice, it prevents people who get one code from using sequential numbers to find others. An additional recommendation (though not implemented here) would be to never delete used codes, just mark them inactive and refuse to display them. In this case, you might want to use a larger random number than 99999999.
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/nilsteampassnet/TeamPass/pull/1015?approve=1'></a>&nbsp;<a href='https://github.com/nilsteampassnet/TeamPass/pull/1015'></a>
<a href='#crh-end'></a>

### DIFF
--- a/sources/items.queries.php
+++ b/sources/items.queries.php
@@ -2710,6 +2710,22 @@ if (isset($_POST['type'])) {
                 DB::delete(prefix_table('otv'), "id=%i", $record['id']);
             }
 
+            // get a list of pre-existing otv ids
+            $existingIDs = array();
+            $rows = DB::query("SELECT id FROM ".prefix_table("otv"));
+            foreach ($rows as $record) {
+                $existingIDs[] = $record['id'];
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/nilsteampassnet/TeamPass/pull/1015'>Refresh</a>.<a href='#crh-update-none'></a>

PR #1265, opened 2016-04-11
Update main.queries.php

diegofau:
Because the data['id'] is empty. The id of user is in $_POST['id']

### DIFF
--- a/sources/main.queries.php
+++ b/sources/main.queries.php
@@ -275,7 +275,7 @@ switch ($_POST['type']) {
                         'ga' => $gaSecretKey
                        ),
                     "id = %i",
-                    $data['id']
+                    $_POST['id']
                 );
 
                 // generate QR url


nilsteampassnet:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/nilsteampassnet/TeamPass/pull/1265'>Refresh</a>.<a href='#crh-update-none'></a>

nilsteampassnet:
No sorry, this doesn't work.
It takes into consideration only on possible call of this case.
$data['id'] has always a value. Take a look to condition at line 237.

Or this solves your particular case, then it means that the root cause is somewhere else.
I will search deeper

diegofau:
I understand the condition is that if the variable $POST [ ' id' ] is empty or not defined will find the id field through consultation, in this case $data['id'] exists , otherwise , if $POST['id'] is completed in the varaible $data will be the login and email but not the id . You can add the field id in the consult sql:
        if (!isset($_POST['id']) || empty($_POST['id'])) {
            // decrypt and retreive data in JSON format
            $dataReceived = prepareExchangedData($_POST['data'], "decode");
            // Prepare variables
            $login = htmlspecialchars_decode($dataReceived['login']);
            $data = DB::queryfirstrow(
                "SELECT id, email
                FROM ".prefix_table("users")."
                WHERE login = %s",
                $login
            );
        } else {
            $data = DB::queryfirstrow(
                "SELECT id, login, email
                FROM ".prefix_table("users")."
                WHERE id = %i",
                $_POST['id']
            );
        }

PR #1482, opened 2016-09-09
Fix personal folder update script for 1476

qubez:
Only attempt resetting personal subfolders if parent exists and didn't
have to be created.

Fixes bug #1476 all folders were set to be private folders if user
without root personal folder existed.


nilsteampassnet:
Thanks

PR #1532, opened 2016-10-05
Added error message in install.js if db-pw contains double quotes 

b13n3rd:
(tooltip for db-password field was not shown in chrome 53 btw)


nilsteampassnet:
Thanks

PR #1553, opened 2016-10-15
Better LDAP support

manusfreedom:
Fix admin setting LDAP help
Add LDAP port
Add support multi LDAP server
Need to add all translation (only english currently)


nilsteampassnet:
Thanks

PR #1556, opened 2016-10-21
Fix bug for upgrading old passwords

jacobsenj:
We tried to update TeamPass and wondered why no password was accessible afterwards. It turned out, that no password has been updated by the upgrade script. This commit fixes the problem.


nilsteampassnet:
Thanks

PR #1560, opened 2016-10-25
Check XFF in addition to REMOTE_ADDR (Fix #1559).

### LINKED ISSUE #1559: API IP Whitelist check does not consider XFF
### Steps to reproduce
1. Deploy TeamPass in environment behind reverse proxy (i.e Docker/Nginx).
2. Set up API whitelist against a real , external source address.
3. Watch API fail with rest_error("IPWHITELIST") because only REMOTE_ADDR is checked.
### Expected behaviour

X-Forwarded-For should also be checked, you should safely assume any webserver or reverse proxy this application is deployed behind properly sanitises this header. 
### Actual behaviour

X-Forwarded-For is not checked, resulting in the user being denied access (as the proxy is seen as REMOTE_ADDR)

I will supply a PR to fix.

lochiiconnectivity:
Also, do not disclose IP in error message if not in whitelist.

Fixes #1559.

### DIFF
--- a/api/functions.php
+++ b/api/functions.php
@@ -17,6 +17,21 @@ $_SESSION['CPM'] = 1;
 require_once "../includes/config/include.php";
 require_once "../sources/main.functions.php";
 
+function get_ip() {
+    if ( function_exists( 'apache_request_headers' ) ) {
+        $headers = apache_request_headers();
+    } else {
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
Thanks

Issue #1218, opened 2016-02-19
Cannot install TeamPass


Guilloma:
On the Server Check section on the Install screen I get this message: 

[ERROR] Answer from server is empty. This may occur if PHP version is not at least 5.5. Please check this this fit your server configuration!

I'm trying to install in debian with PHP 5.6.14 installed. What am I missing?

budicom:
Hi,
Same problem with Ubuntu 14.04 and php 5.5.9.

ymakhloufi:
Hi,

I posted about this very issue several days ago. No solution yet :(
https://github.com/nilsteampassnet/TeamPass/issues/1204

mbartele:
Hi,

I just found a solution for the problem. I answered it in #1105 

I hope this helps you.

Mark

taten14:
I have seen  this error in /var/log/apache2/error.log
Call to undefined function mcrypt_get_iv_size()
So i google it and the answer came, i have to enable mcrypt mod and then restart the apache2 service
php5enmod mcrypt
sudo service apache2 restart

Dante

phfsobhan:
@taten14
Worked for me getting "Answer from server is empty" ( php version error )
Thanks

andresju:
Hi,

I had the same problem. I had everything configured ok but it didn't worked for me. I tried different mysql and php versions and it didn't work for me.

Finally I found a solution that worked for me. I ran the installer in a "Private mode". I think there are any javascript which doesn't work fine and if you run the installer from a private mode or other different browser it has to get all the javascript again and it works.

Andres

FlorentChamault:
Hi 

Do you have another solution for this problem ?
i can't install teampass.. 

thanks

nilsteampassnet:
Did you get through proposed solutions in this ticket? 
Can you provide information such as teampass version, environment, extensions,...

FlorentChamault:
yep, i try all solutions in this ticket but nothing.. 
yep, i can

teampass version => Master on GitHub
environment => debian, apache(Apache/2.4.100, mysql(5.5.50-MariaDB), php (PHP 5.6.29)
extension => 
PHP extension "mcrypt" is loaded 
PHP extension "mbstring" is loaded 
PHP extension "openssl" is loaded 
PHP extension "bcmath" is loaded 
PHP extension "iconv" is loaded 
PHP extension "xml" is loaded 
PHP extension "gd" is loaded 
PHP function "mysqli_fetch_all" is available 
[...]

Issue #1105, opened 2015-11-30
"Cannot redeclare class Encryption\Crypt\aesctr"


H4Mm3r:
Hello,

Since my PHP 5.6 upgrade, I've got this error ont the login POST : 

array(4) {
  ["type"]=>
  int(64)
  ["message"]=>
  string(46) "Cannot redeclare class Encryption\Crypt\aesctr"
  ["file"]=>
  string(64) "/var/www/teampass/includes/libraries/Encryption/Crypt/aesctr.php"
  ["line"]=>
  int(11)
}

Teampass Release : 2.1.19

Thakns by advance

nilsteampassnet:
​What are the steps to repr​oduce this?
What version?

H4Mm3r:
Hi,

Teampass Release : 2.1.19, php 5.6

When connecting I have a 500 on the POST, I have var_dumped the error above...

Thanks

H4Mm3r:
Hello,

I even can't make a fresh install, I've got a 500 on the "Install" POST.

Apache 2.6, PHP 5.6

mbartele:
Have you found any solution?
I am having this issue installing teampass on Ubuntu 14.04.4 and PHP 5.5.9
Other users are also having this problem #1204 and #1218

mbartele:
Hello,

I was facing the same problem today and was able to resolve the issue. I ran into some forum posts stating that in certain versions of PHP, the parent class have to be declared before the extended class.

I edited the "includes\libraries\Encryption\Crypt\aesctr.php" class so that the Aes class is declared before the aesctr class (by just cutting the Aes class part under the aesctr class and pasting it before it).

Unfortunately I have absolutely no PHP experience so I am not able to give a further explanation for this if that is necessary.

I hope I helped some of you out.

Mark

H4Mm3r:
Hi,

Still not working after this change :(

mbartele:
Are you using LDAP?

ymakhloufi:
@mbartele: That worked like a charm!

javimoyabrava:
> Hello,
> 
> I was facing the same problem today and was able to resolve the issue. I ran into some forum posts stating that in certain versions of PHP, the parent class have to be declared before the extended class.
> 
> I edited the "includes\libraries\Encryption\Crypt\aesctr.php" class so that the Aes class is declared before the aesctr class (by just cutting the Aes class part under the aesctr class and pasting it before it).
> 
[...]

PR #1565, opened 2016-10-31
Fix for issues #1218 and #1105.

EddyBeaupre:
This patch fix the the issues #1218 and #1105 but defining the class before its extension.


nilsteampassnet:
Hi

Is the change consisting in moving `class aesctr extends Aes` below `class Aes`?

westerterp:
Yes, I have reviewed the PR and the only change is the moving of the class.

EddyBeaupre:
Yes, that's the only change, defind the class before one derived from it to avoid the error.
Le 29 nov. 2016 6:30 AM, westerterp <notifications@github.com> a écrit :Yes, I have reviewed the PR and the only change is the moving of the class.

Issue #1616, opened 2017-01-12
Bug in IE11: Cannot show password


misterunknown:
### Steps to reproduce
1. Open Teampass in IE11
2. Select a password
3. Click on the little eye symbol: Password shows off for a second and then disappears.

### Expected behaviour
As long as the mouse key is pressed the password should be visible.

### Actual behaviour
The password disappears immediately

### Server configuration
Nothing special

**Web server:**
Apache 2.4

**Database:**
MySQL 5.6

**PHP version:**
PHP 5.6

**Teampass version:**
Teampass 2.1.26

**Updated from an older Teampass or fresh install:**
fresh install

### Client configuration
**Browser:**
IE11

**Operating system:**
Windows 7, Windows 10

nilsteampassnet:
Corrected.

PR #1617, opened 2017-01-12
Fix for issue 1616

misterunknown:
This is a fix for issue #1616. It was tested with Win7/IE11, Win7/FF50, Debian8/FF45, Debian8/Chromium55.


nilsteampassnet:
Thank you

PR #1725, opened 2017-04-08
correction of three fault codes

behnam90:


behnam90:
Fix three bugs.

nilsteampassnet:
Thank you

Issue #1539, opened 2016-10-07
Create Ldap user directly at Teampass DB and  restrict login to Group Ldap  don't work  


andresguisado:
### Steps to reproduce
1. Poxix / OpenLDAP (RFC2307) Search Based - Ldap configuration:
- **Ldap server type:** Poxix / OpenLDAP (RFC2307) Search Based
- **Class to search:** inetOrgPerson
- **User attribute to search:** uid
- **LDAP group to search:** cn=Teampass,ou=Apps,dc=example,dc=com
- **LDAP bind DN:** cn=Manager,dc=example,dc=com
- **LDAP bind password:** XXXX
- **LDAP search base:** ou=People,dc=example,dc=com
- **LDAP domain controller(s):** server01
- **Use LDAP through SSL (LDAPS):** NO
- **Use LDAP through TLS:** NO
- **Teampass local users only:** NO
1. Try to log in and it doesn't work.
### Expected behaviour

Ldap login should work.
1- Teampass should create ldap user in the Database at the first ldap login attempt.
2- Group Ldap filter should work too.If ldap user exist in People OU but doesn’t exist in Teampass Ldap group then ldap login should not be successful however if user exist in both it should be successful.
### Actual behaviour

If ldap user doesn't exist in Teampass DB in advanced, Ldap login doesn't work.

If ldap user exists in Teampass DB in advanced, Ldap login works but group ldap filter doesn’t work.
Regarding to group filter, If ldap user exist in People OU but doesn’t exist in Teampass ldap group, the ldap login is successful too, user should exist in both to be successful.
### Server configuration

**Operating system**:Centos7

**Web server:** Apache 2.4

**Database:** mysql 5.7

**PHP version:**PHP 5.6

**Teampass version:** v.2.1.26 (final release)

**Updated from an older Teampass or fresh install:** Fresh Install 
### Client configuration

**Browser:**Chrome Version 53.0.2785.143 (64-bit)

**Operating system:** Mac 
#### LDAP log

```
* Ldap Log if user doesn't exist in Teampass in advanced, it doesn’t achieve login, never pass login page:

Oct  9 15:58:25 server01 slapd[18870]: conn=1708 fd=30 ACCEPT from IP=X.X.X.X:44194 (IP=0.0.0.0:389)
Oct  9 15:58:25 server01 slapd[18870]: conn=1708 op=0 BIND dn="cn=Manager,dc=example,dc=com" method=128
[...]

PR #1742, opened 2017-04-27
Fix for issue #1539 verifying LDAP groups properly

filo891:
The user was able to log in using LDAP even if he was not in the required group.


nilsteampassnet:
Thank you (I forgot to close)

PR #1808, opened 2017-07-07
Get the database settings from the POST object instead of SESSION

ddrown:
Without this, the "Clean Up" step in the installation fails


nilsteampassnet:
Thank you

PR #1812, opened 2017-07-12
Update #items_list in items.css

mi-cha-el:
Reduced width in #items_list to correct items display under Firefox 54.0


nilsteampassnet:
Fixed in another way

PR #1826, opened 2017-07-25
New dockerfile and docker-compose.yml, docs update

tprobinson:
This approach uses Nginx and more up to date dependencies, as well as converting the images to Alpine. A new Docker Compose example with SSL support is provided, and the docs have been updated on how to use them. Also declutters the repository a little bit by removing apache.conf files.

Let me know if I can help explain anything about it.


nilsteampassnet:
Very good!

Thank you for updating this.

PR #1840, opened 2017-07-27
Update export.queries.php

darrais:
If I have TeamPass installed on a machine with an IP address equal to 255.255.255.255, the link referenced in the "href" attribute (generated when using the offline mode) points to "http://255.255.255.255/teampass/files/...". 
For users with less technological knowledge, what will probably happen is that they'll open the link and write down the generated URL on a paper. In the future, when trying to use it without an internet connection, it won't work.

My suggestion is to add the "download" attribute to the HTML "a" tag created. This attribute allows to download the file referenced in the "href" attribute when clicking on the link. This way, it'll be more clear that a file was created and users need to store it on a USB pen, for example, in order to be able to use it in the future.


nilsteampassnet:
Thank you for the tip.

PR #1851, opened 2017-08-04
Fix LDAP suffix Posix syntax

alemairebe:
I was facing issue with LDAP auth; did a tcpdump to see what was wrong, and I saw a double , in the binddn ; with this fix it works.

### DIFF
--- a/sources/admin.queries.php
+++ b/sources/admin.queries.php
@@ -2049,7 +2049,7 @@ switch ($_POST['type']) {
 
             // Posix style LDAP handles user searches a bit differently
             if ($dataReceived[0]['ldap_type'] === 'posix') {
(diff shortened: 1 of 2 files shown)


nilsteampassnet:
Thank you for your PR.
LDAP is still a mystery for me ... this code (without change) works in this exact condition for a user.
I don't understand why it doesn't on yours ... 
I need to investigate here.

alemairebe:
This is the result of a tcpdump without my patch
ldap fails
This is with the patch
ldap_ok

nilsteampassnet:
I see what I can do to improve all cases.

Thank you for this.

nilsteampassnet:
I did a fix for this

PR #1870, opened 2017-08-16
Update index.php

kmradke:
Logic was backwards.  Javascript was sent when headers were not sent instead of being sent only when headers were already sent.  This can make using a context path for teampass fronted with haproxy difficult.


nilsteampassnet:
Correct.

Thank you.

PR #1873, opened 2017-08-17
Fix session warning in manage_settings -> Custom Fields

jonaseberledmind:
This warning is present in the "Custom Fields" tab due to sending unbuffered data before trying to manipulate headers:

<pre>
Warning: session_start(): Cannot send session cache limiter - headers already sent (output started at /var/www/html/www/admin.settings_categories.php:16) in /var/www/html/www/admin.settings_categories.php on line 20

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/www/admin.settings_categories.php:16) in /var/www/html/www/admin.settings_categories.php on line 37
</pre>


nilsteampassnet:
Thank you

PR #1928, opened 2017-09-26
Fix issue when adding existing item and duplication is not enabled

neilime:
Adding an existing item (label) and duplication is not enabled throws a sql error : Column 'id' cannot be null


nilsteampassnet:
Thank you

PR #1406, opened 2016-07-29
Remove mysqli_fetch_all function dependency (Issue #801)

misterunknown:
Hi,

this fixes #801 .

Cheers,
Marco
<a href='#crh-start'></a><a href='#crh-data-%7B%22processed%22%3A%20%5B%22https%3A//github.com/nilsteampassnet/TeamPass/pull/1406%23issuecomment-245226260%22%2C%20%22https%3A//github.com/nilsteampassnet/TeamPass/pull/1406%23issuecomment-245445107%22%5D%2C%20%22comments%22%3A%20%7B%22General%20Comment%22%3A%20%7B%22html_url%22%3A%20%22https%3A//github.com/nilsteampassnet/TeamPass/pull/1406%23issuecomment-245226260%22%2C%20%22comments%22%3A%20%5B%7B%22body%22%3A%20%22Not%20any%20feedback%3F%22%2C%20%22created_at%22%3A%20%222016-09-07T09%3A29%3A55Z%22%2C%20%22user%22%3A%20%7B%22avatar_url%22%3A%20%22https%3A//avatars.githubusercontent.com/u/3264829%3Fv%3D3%22%2C%20%22html_url%22%3A%20%22https%3A//github.com/misterunknown%22%7D%7D%2C%20%7B%22body%22%3A%20%22My%20feedback%20is%20that%20I%20had%20to%20find%20this%20report%20and%20patch%20to%20even%20install%20the%20software.%22%2C%20%22created_at%22%3A%20%222016-09-07T22%3A58%3A08Z%22%2C%20%22user%22%3A%20%7B%22avatar_url%22%3A%20%22https%3A//avatars.githubusercontent.com/u/1305249%3Fv%3D3%22%2C%20%22html_url%22%3A%20%22https%3A//github.com/lachlan-00%22%7D%7D%5D%2C%20%22title%22%3A%20%22General%20Comment%22%7D%7D%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
- [ ] <a href='#crh-comment-General Comment'></a> &nbsp;<b><a href='https://github.com/nilsteampassnet/TeamPass/pull/1406#issuecomment-245226260'>General Comment</a></b>
- <a href='https://github.com/misterunknown'></a> Not any feedback?
- <a href='https://github.com/lachlan-00'></a> My feedback is that I had to find this report and patch to even install the software.

[...]

PR #1943, opened 2017-10-01
Remove mysqli_fetch_all dependency & update array of routine

DominiqueFuchs:
mysqli_fetch_all still does not seem to be used anywhere throughout teampass. So, pull request #1406 is imho right about removing this dependency (just like the [Official Requirements Page](https://teampass.net/requirements.html) did, it isn't even listed there.

Unfortunately by only removing the specific LOC of the dependency check for this function the check-layout in the routine get's broken because the array is misordered in terms of layout-referencing. This means after removal (like in pull request #1406, which seems to be abandoned and isn't based on actual code in master) the results after the checkpoints partially belong to other lines (example: check for writable /install shows missing extension).

I removed the dependency check for mysqli_fetch_all and reorganized the install.js check-array to reflect the actual order in install.php. Should also be easier to edit or expand in the future this way.


DominiqueFuchs:
Sorry, had to open another pull request #1947 after rebranching.

Issue #1963, opened 2017-10-11
Teampass fails to insert LDAP user trying to log in for the first time into teampass_users table because of null email constraint


leventebakos:
### Steps to reproduce
1. Set up LDAP connectivity in teampass
2. Try to log in with your LDAP connection

### Expected behaviour
A new user should be created in the teampass_users database, and further log on should be enabled for the user.

### Actual behaviour
No new user is created as the mysql insert query in sources/identify.php fails with this error: Query: INSERT INTO `teampass_users` (`login`,`pw`,`email`,`name`,`lastname`,`admin`,`gestionnaire`,`can_manage_all_users`,`personal_folder`,`fonction_id`,`groupes_interdits`,`groupes_visibles`,`last_pw_change`,`user_language`,`encrypted_psk`) VALUES ('login', 'pwd', NULL, NULL, NULL, '0', '0', '0', '0', '', '', '', 1507644625, 'english', '')

Email field has a non-null constraint, causing this error.

### Server configuration
**Operating system**:
CentOS release 6.9 (Final)

**Web server:**
Apache 2.2.15

**Database:**
MySQL 5.1.73

**PHP version:**
PHP 5.6.31

**Teampass version:** 
2.1.27.7

### Client configuration
**Browser:**
Chrome: 61.0.3163.100 (Official Build) (64-bit)

**Operating system:**
Windows 10

#### Web server error log
INSERT INTO `teampass_users` (`login`,`pw`,`email`,`name`,`lastname`,`admin`,`gestionnaire`,`can_manage_all_users`,`personal_folder`,`fonction_id`,`groupes_interdits`,`groupes_visibles`,`last_pw_change`,`user_language`,`encrypted_psk`) VALUES ('login', 'pwd', NULL, NULL, NULL, '0', '0', '0', '0', '', '', '', 1507644625, 'english', '')

leventebakos:
Created pull request:

https://github.com/nilsteampassnet/TeamPass/pull/1964

PR #1964, opened 2017-10-11
Make email field in new LDAP user insertion null safe

leventebakos:
Fix for issue "Teampass fails to insert LDAP user trying to log in for the first time into teampass_users table because of null email constraint".

https://github.com/nilsteampassnet/TeamPass/issues/1963


nilsteampassnet:
Thank you

PR #1979, opened 2017-10-25
translated some keys

sarperkaya:


nilsteampassnet:
Please use poeditor.com/projects/view?id=16418 to improve strings.

PR #1988, opened 2017-10-30
Pragma: no-cache is a request header, it has no meaning in response

chipitsine:
according to https://tools.ietf.org/html/rfc7234 Pragma: no-cache
is intended to use in request

it is common for php however to add Pragma: no-cache to response
due to bad documentation example http://mx1.php.net/manual/ru/function.header.php


nilsteampassnet:
Thank you

PR #2030, opened 2017-11-22
Add Checkbox as a Category Item type

atplack-tribridge:
This change allows Categories to have custom Items which are a checkbox.  Checkbox items are good for values that should remain boolean and will allow API interfaces to be more consistent.

This change required the addition of an additional element in the language files.  I have updated all the language files and noticed two changes in the Japanese and Ukranian files.  I am thinking this is due to Unicode differences on my laptop.  I would suggest a review of these two files and correction of these elements if needed before the pull is accepted.


atplack-tribridge:
Found a couple of issues, will resubmit.

PR #2031, opened 2017-11-23
Add checkbox as a category custom item

atplack-tribridge:
This pull modifies the code to create a new category custom item type which is a checkbox.  The checkbox is a cleaner way of dealing with boolean values when interfacing with the API.

Because the database stores this value as a string, the code takes care to convert between string and the boolean value of the checkbox.

To add the string values, the language files needed updating.  I have attempted to get the French as close as possible, but the other languages are in English.  In developing the code, there were two language files which "updated" but were not a change by me.  The Unicode values look correct now as the file was copied from a fresh git pull.  However, GitHub is still flagging these values as different.  I have noted these in the internal comments of this pull request.


includes/language/ukrainian.php
```diff
@@ -41,7 +43,7 @@
     'email_ga_subject' => 'Ваш Google Authenticator флеш код для Teampass',
     'email_ga_text' => 'Вітаю, <a href=\'#link#\'>перейдіть за цим посиланням</a> і зіскануйте QR код за допомогою застосунка Google Authenticator. Це потрибно для налаштування вашого флеш-коду',
     'settings_attachments_encryption' => 'Ввімкнути шифрування для вкладень',
-    'settings_attachments_encryption_tip' => 'ЦЕ НАЛАШТУВАННЯ МОЖЕ ПОШКОДИТИ ІСНУЮЧІ ВКЛАДЕНН�',
+    'settings_attachments_encryption_tip' => 'Ð¦Ð• ÐÐÐ›ÐÐ¨Ð¢Ð£Ð’ÐÐÐÐ¯ ÐœÐžÐ–Ð• ÐŸÐžÐ¨ÐšÐžÐ”Ð˜Ð¢Ð˜ Ð†Ð¡ÐÐ£Ð®Ð§Ð† Ð’ÐšÐ›ÐÐ”Ð•ÐÐÐ',
```
atplack-tribridge: As best as I can tell, this Unicode value was not changed by my code, but still shows as different, I have updated the value to the exact same string in GitHub directly, however, this difference still shows.

includes/language/japanese.php
```diff
@@ -52,7 +54,7 @@
     'admin_action_pw_prefix_correct_tip' => 'Please backup the database before performing this task. This script will perform an update of password prefix in database. Only run this to fix incorrectly displayed passwords after upgrading from older versions.',
     'items_changed' => '変更しました。',
     'ga_not_yet_synchronized' => 'Get identified with Google Authenticator',
-    'ga_scan_url' => 'Google認証アプリなどでこのQRコードをスキャンしてください�',
+    'ga_scan_url' => 'Google隱崎ｨｼ繧｢繝励Μ縺ｪ縺ｩ縺ｧ縺薙�QR繧ｳ繝ｼ繝峨ｒ繧ｹ繧ｭ繝｣繝ｳ縺励※縺上□縺輔＞',
```
atplack-tribridge: As best as I can tell, this Unicode value was not changed by my code, but still shows as different, I have updated the value to the exact same string in GitHub directly, however, this difference still shows.

Issue #2029, opened 2017-11-22
Another Error on Step 3 of Installation


jdipper:
### Steps to reproduce
1. Have keys in the _install table
2. Try to complete step 3.

### Expected behaviour
Database should be updated with new values.

### Actual behaviour
Catchable fatal error: Argument 3 passed to mysqli_query() must be an instance of int, string given in /var/www/teampass/install/install.queries.php on line 232

### Server configuration
**Operating system**:
Centos

**Web server:**
H2O

**Database:**
MariaDB

**PHP version:**
5.6.99-hhvm

**Teampass version:** 
2.1.27.10

**Teampass configuration file:**
Not generated

**Updated from an older Teampass or fresh install:**
Fresh install

### Client configuration
**Browser:**
Chrome

**Operating system:**
OSX

#### Web server error log
```
Catchable fatal error: Argument 3 passed to mysqli_query() must be an instance of int, string given in /var/www/teampass/install/install.queries.php on line 234
```

#### Log from the web-browser developer console (CTRL + SHIFT + i)
```
jquery.min.js:18 POST http://10.1.4.100/install/install.queries.php 500 (hphp_invoke)
send @ jquery.min.js:18
ajax @ jquery.min.js:18
checkPage @ install.js:183
onclick @ install.php:71
install.js:200 Uncaught TypeError: Cannot read property '0' of null
    at Object.complete (install.js:200)
    at Object.resolveWith (jquery.min.js:16)
    at w (jquery.min.js:18)
    at XMLHttpRequest.d (jquery.min.js:18)
```

#### Notes

So I guess my previous error message led me to this.  You have an error with how mysqli_query is called if data already exists on lines 232 and 238:

`mysqli_query($dbTmp, "UPDATE `_install` SET `value` = '", empty($session_url_path) ? $db['url_path'] : $session_url_path, "' WHERE `key` = 'url_path';");
`
I suspect it should be:
`mysqli_query($dbTmp, "UPDATE `_install` SET `value` = '". empty($session_url_path) ? $db['url_path'] : $session_url_path ."' WHERE `key` = 'url_path';");
`
ie full stops / periods rather than commas, otherwise `empty($session_url_path) ? $db['url_path'] : $session_url_path` is being submitted as the 3rd argument for mysqli_query and `' WHERE `key` = 'url_path';` as the 4th query.

jdipper:
Just tested it, and that fixed the problem for me, and I also put the ternary operator in brackets.

[...]

PR #2034, opened 2017-11-23
Fixes to Step 3 of install

jdipper:
Hi!

I noticed that #2029 only fixed on of the lines, not the second (line 232).  Also, $db was being used for url_path and abspath rather than $data.

This PR fixes these problems.

Regards,

Jack


nilsteampassnet:
Yes you are right.

Thank you

PR #2111, opened 2018-02-03
Add support for login through http header.

finalbeta:
Hi Nils, 

I've created a pull request to add support for "BASIC AUTH" (HTTP header login) automatic logon into teampass. 

This adds one option to the admin panel to enable the feature, yes or no. 
If enabled, and the login information is available in the http header, this will show the login name on the login form (read only), it will hide the password field. 
The logon process will then start after 3 seconds. 
The logon process then grabs the information from PHP (not the request) to log the user in to the system.
Two language variables have been added. (only for english at the moment)
Maintenance mode disables everything. 

I would love to see this land in teampass, I'm open to suggestions.

### DIFF
--- a/admin.settings.php
+++ b/admin.settings.php
@@ -198,6 +198,18 @@ echo '
              </td>
             </tr>';
 echo '<tr><td colspan="3"><hr /></td></tr>';
+// Use authentication information in HTTP header
+echo '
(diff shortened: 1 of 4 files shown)


nilsteampassnet:
merged.

Thank you

PR #2230, opened 2018-04-25
API support for file attachments

opkorpela:
API support for uploading, listing, and downloading file attachments to password items.


opkorpela:
Api calls for adding, listing and downloading attachment files to password items.

I merged this to development branch.

PR #2249, opened 2018-05-07
Item add/update separate parameter encoding

opkorpela:
Api change for adding and updating items. The parameters are base64-encoded separately. This prevents problems caused by semicolons inside item description.


nilsteampassnet:
would it be possible to prepare the PR on the 'development' branch please?

opkorpela:
The changes are now based on the development-branch.

opkorpela:
I also changed the api so, that the old api won't break. The new item adding/updating calls are marked as "version 2".

PR #2265, opened 2018-05-15
Version 2 api calls for creating and updating items

opkorpela:
This pull request is for fixing the problem with the semicolons inside the item label, password, description, or login. For example, if you try to add a new item with a semicolon inside the description text, the semicolon is handled as a parameter field separator. The current way of  base64-encoding all the parameters together doesn't resolve this problem.

### DIFF
--- a/api/functions.php
+++ b/api/functions.php
@@ -892,6 +892,146 @@ function rest_get()
                 } else {
                     rest_error('ITEMMISSINGDATA');
                 }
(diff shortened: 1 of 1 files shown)


mrlocke:
When this will be merged? Hit this problem today on 2.1.27.16 and was about to submit a patch, when I've seen this.
Hope this will go into a new version.

nilsteampassnet:
Thank you for this.
I do not do a merge but took from your PR the main change.
Included in API 2.1.2

Issue #2314, opened 2018-06-22
SQL error in API near user name


mayeulk:
The following API call returns an error:                    

wget -O saved_pw_teampass.json http://127.0.0.1/teampass/api/index.php/read/userpw/myuser?apikey=myAPIkeyXXXX24NPwwzzxvtM9F77LmvehDrNA
cat saved_pw_teampass.json              
The error:
  "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''myuser''' at line 3B/wsWwsLCw[shortened]"
  Note the 2 opening single quotes and the 3 closing single quotes.
The bug seems to be an extra single quote line 418 of functions.php, see:
https://github.com/nilsteampassnet/TeamPass/blob/454057eb49c52ac885e32fb3810e402b9019f8a4/api/functions.php#L418

mayeulk:
Patch here: https://github.com/mayeulk/TeamPass/blob/master/api/functions.php

nilsteampassnet:
Thank you

PR #2315, opened 2018-06-22
Remove single quote in SQL for API call (near user name)

mayeulk:
Fixes https://github.com/nilsteampassnet/TeamPass/issues/2314


nilsteampassnet:
Thank you

PR #2346, opened 2018-07-16
[CSV-Import] convert field to string bevore using replace()

borzel:
## Description
If you import this csv-file [demo-cvs.txt](https://github.com/nilsteampassnet/TeamPass/files/2198685/demo-cvs.txt) the import does stop and shows waiting:

grafik

## Error
grafik

## Cause

Sometimes the values of the csv are interpreted as "number" and than the replace()-function fails, because it does not work on numbers.

grafik

## Fix
I fixed this with converting the variable `string` to string. With this fix the demo-CSV can be imported.

### DIFF
--- a/includes/js/functions.js
+++ b/includes/js/functions.js
@@ -181,8 +181,8 @@ function unsanitizeString(string){
 **/
 function sanitizeString(string){
     if(string !== "" && string !== null && string !== undefined) {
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
This is correct.

Thank you

PR #2354, opened 2018-07-27
Api read userfolders json fix

opkorpela:
There was a syntax error in api call read/userfolders. The variable $json should be an array, not string.


nilsteampassnet:
Thank you

PR #2355, opened 2018-07-27
Return the parent folder ids on API call read/userfolders

opkorpela:
Added the parent folder id to the return values.


nilsteampassnet:
Thank you

PR #2362, opened 2018-08-03
Removed excess item id from API url add/item

opkorpela:
I made a fix to the add/item API url and fixed the urlSafeB64Decode function calls.


nilsteampassnet:
Thank you

PR #2369, opened 2018-08-08
Quickly support POST, working around a security vulnerability

rseabra:
Quickly support POST, working around a security vulnerability that would leave passwords easily discoverable on a web server's logs.

The API has everything as a GET including **writing** new passwords. This patch very quickly adds support to POST.

Even if not accepted, please fix this issue as it's a very bad issue. BASE64 on top of values hides nothing, really.


nilsteampassnet:
Idea merged

PR #2396, opened 2018-08-28
Fix login when AD LDAP with non-expiring user passwords is used

jannero:
If the user is disabled due to password expired, LDAP bind does
not succeed, i.e. the removed code would never be effective.

You could fix this also by handling all return values from passwordExpiry but I see no reason as the LDAP bind fails when the user's password is expired or when the user is inactive (expired).

### DIFF
--- a/sources/identify.php
+++ b/sources/identify.php
@@ -761,12 +761,6 @@ function identifyUser(
                     $ldapConnection = true;
                 }
 
-                // Is user expired?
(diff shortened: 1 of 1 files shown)


jannero:
@nilsteampassnet So are you planning to fix this issue in some way?

As I said in the PR's description you can also fix the issue by handling all the real return values of _passwordExpiry_ function if you don't want to get rid of that part of the code for some reason.

PR #2431, opened 2018-10-10
Correct item creation/edit tab label

izomorph:
There is no visibility settings in item dialog.
Although tab label too long and breaks the tab layout.


nilsteampassnet:
Thanks

Issue #2100, opened 2018-01-22
API not working in container.


lcworld:
Hello

### Steps to reproduce
1. download docker-compose.yml from this folder
2. execute docker-compose up
3. configure teampass, and disable maintenance mode
4. enable API and create a key.
5. open http;//teampass/api/index.php/info/version/?apikey=...  (replace ... with key genreated in 4
    It returns HTTP error 404.

It's the same for all other items of the API.

### Expected behaviour
It should return the version in json format.

### Actual behaviour
It returns HTTP error 404

### Server configuration
Docker containers generated with docker-compose

### Client configuration
**Browser:**
All / curl in command line
**Operating system:**
All

#### Web server error log
```
172.18.0.2 - - [22/Jan/2018:23:33:28 +0000] "GET /api/index.php/info/version/?apikey=... HTTP/1.1" 404 1220 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36"
```

How to solve the issue ?

After some search in internet, I'm wondering if a URL rewriting is not missing in the nginx server configuration ?

stravassac:
@lcworld like this thread
https://github.com/nilsteampassnet/TeamPass/issues/566
you can add in /etc/nginx/sites-available/default.conf
         location /api/ {
                try_files $uri $uri/ /api/index.php?$args;
       }
and in bash kill -HUP 1 (to reload)

CayoM:
just for the record: this change doesn't seem to be in the latest docker image  from docker hub.

Anyways this doesn't seem to work for me. I keep getting a 301:

```
curl -k -XGET http://localhost:8088/api/index.php/read/folder/1?apikey=AxSdxVR76pn9P2QSYm8h6mjNAQg6sYvrDS3WXnU
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

as mentioned above, i updated the nginx config file:
```
        location /api/ {
                try_files $uri $uri/ /api/index.php?$args;
        }
```

```
[...]

Issue #566, opened 2014-05-30
Teampass API and nginx


brantje:
On nginx when i have enabled the api, and i access the api i get the
 404 not found error when opening index.php/read/category/0
Using nginx /1.4.7

```
var_dump($GLOBALS['request']);
array(1) {
  [0]=>
  string(0) ""
}
```

Do i need some add some line in nginx.conf?
For drupal i have to add this one

```
        location / {
                try_files $uri $uri/ /index.php?$args;
       }
```

Great that you add an API, maybe expand it abit more so users can login with the api?
I already started some designs for an android app =).

PR #2452, opened 2018-10-27
Fix API URL

bezigeboefjes:
BUG: API not working in container. #2100
https://github.com/nilsteampassnet/TeamPass/issues/2100

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,13 @@ RUN echo && \
   echo "max_execution_time = 120" >> /usr/local/etc/php/conf.d/docker-vars.ini && \
 echo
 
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
Question, the change shows the variable `$uri` but where is it set?

bezigeboefjes:
> Question, the change shows the variable `$uri` but where is it set?

That is in the Nginx configuration. I've based my fix on the comment:
--
@lcworld like this thread
#566
you can add in /etc/nginx/sites-available/default.conf
location /api/ {
try_files $uri $uri/ /api/index.php?$args;
}
and in bash kill -HUP 1 (to reload)
--

PR #2468, opened 2018-11-07
Wrong path

krustophenia:
Hi.

In the teampass-docker-start.sh the path where the repo is being cloned is wrong. 
Right now, after start the docker-compose you need to get inside the teampass container and move everything from /var/www/html/www to /var/www/html.
Now it shows:
git clone $REPO_URL ${VOL}/www
and it should be:
git clone $REPO_URL ${VOL}


fheuer:
I figured out that changes in Dockerfile and teampass-docker-start.sh are necessary to fix the issue stated above as well as in #2472 

PR incoming

Issue #2472, opened 2018-11-11
The nginx webroot is incorrect in the latest teampass docker container


Andrei-Paul:
### Steps to reproduce
1. Replace the old docker image with the new one.
2. Start the container.
3a. Try the https://normal.url.
3b. Try the https://normal.url/sk/teampass-seckey.txt.

### Expected behaviour
3a. Get the login page.
3b. Get file not found page.

### Actual behaviour
3a. Get 403 error.
3b. Get sk/teampass-seckey.txt

### Server configuration
**Operating system**:
Latest Docker Container - a1c5a2860068

**Web server:**
Default NginX instance in Teampass Container

**Teampass version:** 
TeamPass 2.1.27.23

----------------------------------------------------------------
The current docker image has nginx /etc/nginx/sites-enabled/default.conf root like:
root /var/www/html;
instead of:
root /var/www/html/www;

/etc/nginx/sites-available/default.conf is still correct for some reason.

Also, this allows access to the /var/www/html/sk/whatever because it is now in the web root ...

Andrei-Paul:
In https://hub.docker.com/r/teampass/teampass/, at:
[...]
With Docker Compose
[...]
Note: Use /var/www/html/sk as your "Absolute path to saltkey" during installation.

In hindsight, this was probably not a good idea ...

axnsan12:
This is caused by https://github.com/nilsteampassnet/TeamPass/commit/84f3da5fd600e5204ec1de09b1434a2d1ff0231a which breaks the ``sites-enabled`` symlink to ``sites-available`` by using ``sed -i``.

The ``nginx-php-fpm`` scripts edit only the ``sites-available`` configs: https://github.com/richarvey/nginx-php-fpm/blob/584b3272e9234c99e8cdbbd018350ef7e76a5ebc/scripts/start.sh#L21-L23

Should probably use ``--follow-symlinks`` on the ``sed``.

fheuer:
Can confirm the issue.

nilsteampassnet:
Thank you for reporting this.
As you can see, I'm not an expert in performing for images.
But I will improve this.

nilsteampassnet:
Please reopen this ticket on need.

lsascha:
Sorry to write here, i tried to install it on a docker swarm, but i get only a phpinfo when accessing the domain using port 80.

i did not add the ngingx proxy since the swarm already uses traefik as proxy for container services.

any idea? Am i configuring it wrong?

PR #2476, opened 2018-11-16
nginx webroot fix 

fheuer:
Fix for #2472


fheuer:
can be ignored ; has been replaced by pr #2477

PR #2477, opened 2018-11-16
fix for #2472

fheuer:
changes on Dockerfile to build image with proper environment variable webroot
changes in start.sh to clone to correct directory and work on it


nilsteampassnet:
Thank you @fheuer 

I will merge this inside `development` branch

Issue #2486, opened 2018-11-30
Editing roles on users wipes dissimilar roles


IanRaceASP:
### Steps to reproduce
1. Create three permissions roles (A,B,C) and two users (1,2) like so:
    User 1 (manager): Member of role A, B
    User 2 (non-manager): Member of role B, C
2. As User 1, modify the roles of User 2. Give them role A.

### Expected behaviour
User 2 ends up being a member of roles A, B and C.

### Actual behaviour
User 2 ends up being a member of roles A and B only, their membership in role C has been wiped because the manager was not a member of role C and therefore couldn't see role C in the roles editing dialog.

### Additional information

We were able to resolve this issue (although in a workaround sort of manner) by adding a bit to a line at `teampass/sources/users.queries.php`, PR to come shortly...

### Server configuration
**Operating system:** Linux 4.9.0-7-amd64 #1 SMP Debian 4.9.110-3+deb9u2 (2018-08-13) x86_64 GNU/Linux

**Web server:** Apache/2.4.25 (Debian)

**Database:** mySQL

**PHP version:** PHP 7.0.30-0+deb9u1

**Teampass version:** 2.1.27.22

**Teampass configuration file:** https://gist.github.com/IanRaceASP/1021c41e8638ff1f1b94a382b3e8b4aa

**Updated from an older Teampass or fresh install:** Fresh install

### Client configuration
**Browser:** Firefox 63.0.3 (64-bit)

**Operating system:** Windows 10 Home 1809 (18267.1001)

IanRaceASP:
PR at https://github.com/nilsteampassnet/TeamPass/pull/2487

nilsteampassnet:
Yes this is absolutely right.

Thank you for your PR.

IanRaceASP:
Thank you for your work on TeamPass!

PR #2487, opened 2018-11-30
Workaround for https://github.com/nilsteampassnet/TeamPass/issues/2486

IanRaceASP:
Makes it so that the roles displayed in the role editing dialog for a manager will include the roles that the other party is a member of, which the manager is _not_ a member of. This avoids wiping the other roles of a user by omission.


nilsteampassnet:
Thank you

Merged

PR #2503, opened 2018-12-19
Update Dockerfile to use $uri and $args variables

CayoM:
Since the Docker-Containers were having trouble with the API, the provided solution (adding the /api/ location to NGINX) solves the problem, however it was implemented wrong. Those variables have to be escaped in the Dockerfile, otherwise the value of the variable will be added (in this case a empty value).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ echo
 # Search last } and insert configuration rows before
 RUN sed -i "/^}/i \
   location /api/ {\
-          try_files $uri $uri/ /api/index.php?$args;\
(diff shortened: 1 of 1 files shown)


CayoM:
any chance this will get implemented? The current Dockerfile from official repo is broken for API usage.

CayoM:
bump

nilsteampassnet:
Thank you.

I will merge it.

CayoM:
The code was merged, but there is no new image on dockerhub (the image is still 4 months old). How is the automated dockerbuild triggered?

ricain59:
Hi,

You can update the docker please? 

Thank you.

PR #2505, opened 2018-12-19
Update readme.md

rokx:
Without mounting /var/php/session my installation freezed. Included sample docker run command.


nilsteampassnet:
Will merge it.

Thank you

PR #2560, opened 2019-03-01
Fix an issue for one time password changes if current user is not root

wanghaifeng:
If the login user is not root, type in the current password firstly, then type in the new password.


nilsteampassnet:
Tahnk you

I will merge it

Issue #2936, opened 2021-06-29
LDAP Authentication fails if using sAMAccountName


hitenmandalia:
### Steps to reproduce
1. Configure username for connecting to AD with sAMAccountname 
2. User name attribute set to sAMAccountName
3. User Object Filter set to (objectCategory=Person)(sAMAccountName=*))
4. Test configuration with sAMAccountName of a user

### Expected behaviour
Authentication succeeds

### Actual behaviour
Caution
Error : 49 - Invalid credentials
80090308: LdapErr: DSID-0C090453, comment: AcceptSecurityContext error, data 52e, v3839

### Server configuration
**Operating system**:
Alpine Linux

**Web server:**
Apache

**Database:**
Amazon Aurora RDS (mysql 5.7)

**PHP version:**
7.4

**Teampass version:** 
3.0.0.8

**Teampass configuration file:**

**Updated from an older Teampass or fresh install:**
PLEASE attach to this issue the file `/includes/config/tp.config.php`.

### Client configuration
**Browser:**

**Operating system:**

#### Web server error log
```
No Errors in Log
```
#### Log from the web-browser developer console (CTRL + SHIFT + i)
```
Insert the log here and especially the answer of the query that failed.
```

nilsteampassnet:
Hello,
I have similar setup and it works smoothly.

![Capture d’écran de 2021-07-03 09-12-03](https://user-images.githubusercontent.com/1197546/124348269-449dfe80-dbe9-11eb-85f9-38547f2a3b2a.png)

While googling your error, I found this page https://ldapwiki.com/wiki/Common%20Active%20Directory%20Bind%20Errors
Perhaps this could help you but I don't see how the Adldap could be wrong there.
I would be greatly interested in having any feedback.
Perhaps something needs to be robustified.

hitenmandalia:
![image](https://user-images.githubusercontent.com/62652225/124630963-d2dce380-de7a-11eb-91e0-d45be38ec244.png)

This is the error I keep getting.   I even get the same error when testing and using my own credentials which I know work.

hitenmandalia:
weird thing.  If I use cn as the username with which to bind, change the username attribute to cn and then test the configuration with the following:

<firstname> <surname>  (theres a space between the two), it succesfully passes.  as soon as I change back to samaccountname it fails.  @nilsteampassnet  the only difference i can see between both of our configs is that you are using OpenLDAP whereas I am using Active Directory.  

[...]

PR #2964, opened 2021-09-06
Fix: AD LDAP connection and users/groups fetching (#2936)

relkai:
I stumbled across the same problem as in issue #2936 and worked on a solution.

This PR solves not only the problem of the failing connection test, but also the non-functioning LDAP synchronization of AD users and groups.
In addition, the array containing the AD groups (for role mapping) is sorted alphabetically for a better overview.
This is especially helpful in large AD environments with a lot of groups.

I'm not sure, if this breaks other LDAP implementations, so please carefully test this PR.


relkai:
What the heck...??
Sorry, I think I completely messed up this PR. I will close and re-open it.

PR #2965, opened 2021-09-06
Fix: AD LDAP connection and users/groups fetching (#2936)

relkai:
I stumbled across the same problem as in issue #2936 and worked on a solution.

This PR solves not only the problem of the failing connection test, but also the non-functioning LDAP synchronization of AD users and groups.
In addition, the array containing the AD groups (for role mapping) is sorted alphabetically for a better overview.
This is especially helpful in large AD environments with a lot of groups.

I'm not sure, if this breaks other LDAP implementations, so please carefully test this PR.

### DIFF
--- a/sources/ldap.queries.php
+++ b/sources/ldap.queries.php
@@ -106,7 +106,7 @@ switch ($post_type) {
         $config = [
             // Mandatory Configuration Options
             'hosts'            => [explode(",", $SETTINGS['ldap_hosts'])],
-            'base_dn'          => $SETTINGS['ldap_bdn'],
+            'base_dn'          => (isset($SETTINGS['ldap_dn_additional_user_dn']) ? $SETTINGS['ldap_dn_additional_user_dn'].',' : '').$SETTINGS['ldap_bdn'],
             'username'         => $SETTINGS['ldap_username'],
             'password'         => $SETTINGS['ldap_password'],
         
@@ -180,13 +180,12 @@ switch ($post_type) {
         }
 
         // User try to auth
-        $connection->query()
+        $userLogin = $connection->query()
             ->where($SETTINGS['ldap_user_attribute'], '=', $post_username)
             ->firstOrFail();
 
         try {
-            $connection->auth()->bind($SETTINGS['ldap_user_attribute'].'='.$post_username.','.(isset($SETTINGS['ldap_dn_additional_user_dn']) ? $SETTINGS['ldap_dn_additional_user_dn'].',' : '').$SETTINGS['ldap_bdn'], $post_password);
-
+            $connection->auth()->bind($userLogin[$SETTINGS['ldap_user_attribute']][0], $post_password);
         } catch (\LdapRecord\Auth\BindException $e) {
             $error = $e->getDetailedError();
             
--- a/sources/main.functions.php
+++ b/sources/main.functions.php
@@ -2796,7 +2796,7 @@ function ldapCheckUserPassword(string $login, string $password, array $SETTINGS)
     $config = [
         // Mandatory Configuration Options
         'hosts' => [$SETTINGS['ldap_hosts']],
-        'base_dn' => $SETTINGS['ldap_bdn'],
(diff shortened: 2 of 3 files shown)


relkai:
AD login is still not possible, and obviously I'm still missing an essential part but I can't find it.
Any hints, where else to look at, if it's not the function `ldapCheckUserPassword` (which should be fixed with the latest commit)?

liedekef:
I can testify that this pull request is needed. Reason: you can't expect to know how a DN of a user in ldap is built. From the original code it is assumed to be something like uid=login,dc=my,dc=site (just an example). But in many installations the DN is more cn=myname,dc=my,dc=site. So there's always the need to first search the user, get the DN from the resulting entry and then use that to login.
Also, it seems that sAMAccountName needs to be lowercased in order to work correctly. And this pull needs other fixes done later in time to actually make ldap logins work.

liedekef:
Edit, next to this pull request, ./sources/identify.php needs to be fixed too, in the function authenticateThroughAD:
`        'username'         => $SETTINGS['ldap_user_attribute']."=".$username.",".(isset($SETTINGS['ldap_dn_additional_user_dn']) ? $SETTINGS['ldap_dn_additional_user_dn'].',' : '').$SETTINGS['ldap_bdn'],`

Next to that, in several places there is now the option:
LDAP_OPT_X_TLS_REQUIRE_CERT => LDAP_OPT_X_TLS_HARD

which rarely works ok for internal AD servers. The best is in fact to ignore the cert:
LDAP_OPT_X_TLS_REQUIRE_CERT => LDAP_OPT_X_TLS_NEVER

(that's a change I also needed)

relkai:
Thank you @liedekef 

I tried to fix this file too, but just can't get AD logins to work properly.
If I see it correctly, this script tries to directly connect to AD with the entered credentials, which potentially could work, but seems to be wrong by design.
So similar to my previous commits I changed it to use the configured `$SETTINGS['ldap_username']` and `$SETTINGS['ldap_password']` for the initial connection, do a search for the entered username and use the DN together with `$passwordClear` for authentication.
I also changed `LDAP_OPT_X_TLS_HARD` to `LDAP_OPT_X_TLS_NEVER` (although this didn't cause any problems in my environment in the other functions) but regardless what I do, I can't log into TeamPass using my credentials.

Did you manage to get it work on your side?

liedekef:
Yes I did. But due to the code used, you need to use "samaccountname" as attribute. Using uppercase or a mix won't work as expected. I currently have no access anymore (weekend), but I'll check other changes on Monday. I remember not doing the bind-search and just using username@extension (forced the extension in that php file) to be quick (on AD you can authenticate like that too) and then it all just worked (but your change, searching for the DN is the better way to go).
Based on what I see, the current code is not ready for generic LDAP use just yet, but with your change it's getting there :-)

relkai:
Aaah yes, that would also do the trick.
But as you already pointed out, a generic solution would be better. Otherwise each user needs to adjust the code according to the environment used.

By the way - I use `mail` instead of `sAMAccountname` for authentication, because our users are used to it in other portals. So this should also work straight away, but there must be something I'm missing, because it's still failing.
Strange thing is, that it's working with my changes in the connection test and users fetching now. But the same changes adopted to identify.php are failing.

I will keep on playing around with the code, but would be cool to take a look at your changes (no hurry, please).

liedekef:
I verified all other changes and found none. So just identify.php needs to be corrected. I haven't seen your code change for that, but I'm willing to play around with it if you'd like

FooBarTrixibell:
I think the design is not wrong, BindDN should not be needed during authentication as you are passing what should be a valid AD user who can (if the details are correct) read their own user attributes. I think I recall some vulnerability caused by using bind details in auth somewhere and being told it was bad practice for anything but sync.

I also suspect that 

`LDAP_OPT_X_TLS_REQUIRE_CERT => LDAP_OPT_X_TLS_HARD,`

should only be set if 

`'use_tls' => $SETTINGS['ldap_tls'] === 1 ? true : false,`

resolves to being true.

liedekef:
Many LDAP servers don't allow unauthenticated searches, so you need an initial user that can do lookups. It is not just AD here, but any general LDAP server needs to be considered.

PR #2999, opened 2022-01-08
Teampass 3.0 fix Active Directory

manusfreedom:
Fix Active Directory authentication (main & test)
Fix Active Directory users list
Fix missing string convert to int for $SETTINGS
Fix typo in user_sharekeys_reencryption_next
Replace user_password_not_correct by password_is_not_correct

### DIFF
--- a/error.php
+++ b/error.php
@@ -79,7 +79,7 @@ if (
     }
 
     //Log into DB the user's disconnection
-    if (isset($SETTINGS['log_connections']) && $SETTINGS['log_connections'] === 1) {
+    if (isset($SETTINGS['log_connections']) && (int) $SETTINGS['log_connections'] === 1) {
         logEvents($SETTINGS, 'user_connection', 'disconnection', (string) $_SESSION['user_id'], $_SESSION['login']);
     }
 } else {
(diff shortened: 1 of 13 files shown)


manusfreedom:
Not fixed:
* When adding LDAP users from AD => `Generating keys` remains and does not turn off, but in the JS console logs we get `Performing finished` and `Generation des clés terminée` correctly.
* Your temporary encryption code does not work, currently we need to put your current AD password.
Other than that, it seems to work.

Used LDAP config:
* Hosts: `mydc.mydomain.com`
* LDAP Port: `636`
* Base DN: `DC=mydomain,DC=com`
* Username: `CN=Teampass,OU=Services,OU=Global,DC=mydomain,DC=com`
* Password: `*********`
* User name attribute: `userprincipalname`
* Additional User DN: `OU=Users,OU=Global`
* User Object Filter: `(&(objectCategory=Person)(userPrincipalName=*))`
* LDAP server type: `Active Directory`
* Use LDAP through SSL (LDAPS): `Yes`
* Use LDAP through TLS: `Yes`
* Teampass local users only: `No`
* Local and LDAP users: `No`
* Hide forgot password link on Home page: `Yes`
* Username to use for testing purpose: `myuser@mydomain.com`
* Username password: `*********`

nilsteampassnet:
Thank you @manusfreedom

PR #3021, opened 2022-01-30
More explicit temporary code e-mail

manusfreedom:


nilsteampassnet:
Thank you @manusfreedom 

Nevertheless please use https://poeditor.com/projects/view?id=433631 for translation purpose ;)
It makes me double work when done through PR/

Thanks anyway

PR #3024, opened 2022-01-31
Fix view icon works one time only

manusfreedom:


nilsteampassnet:
Thank you @manusfreedom

PR #3025, opened 2022-01-31
Fix error 500 when logItems is called

manusfreedom:


nilsteampassnet:
Thank you @manusfreedom

PR #3027, opened 2022-02-01
Fix error in admin to display logs

manusfreedom:
* Fix error with stripslashes and null value
* Fix logs sorting order


nilsteampassnet:
Thank you @manusfreedom

PR #3084, opened 2022-03-15
Update options.php (Maintenance Mode ON), otv.php, index.php (One-Time View doesn't work)

dmotelica:
Maintenance mode always ON bug : options.php 
One-Time View doesn't work : otv.php, index.php


nilsteampassnet:
Thank you

PR #3085, opened 2022-03-15
Update nginx image to supported PHP version

ImNtReal:
This PR will update the nginx image used to one which has a supported version of PHP. It also adds support for specifying a specific git tag when performing the initial clone.


nilsteampassnet:
Thank you

PR #3118, opened 2022-04-04
Use user provided LDAP filter (#3040)

### LINKED ISSUE #3040: LDAP sync ignores User Object Filter
### Expected behaviour
After setting up LDAP and subsequent synchronization via LDAP, the User Object Filter is ignored.
Since there are a lot of objects in our domain, after 3-6 minutes the domain controller returns the result of the request, but we get a PHP error (memory_limit = 512M) _(url is intentionally hidden)_:
```
PHP Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /var/www/teampass/includes/libraries/LdapRecord/Ldap.php on line 18, referer: https://url/index.php?page=users
```
As a result, synchronization does not occur.

### Actual behaviour
Return objects (correct searchRequest) according to the User Object Filter (in this case, only those objects that are in a specific group)

### Server configuration
**Operating system**:
Debian GNU/Linux 11 (bullseye)

**Web server:**
Apache/2.4.52 (Debian)

**Database:**
10.5.12-MariaDB-0+deb11u1 Debian 11

**PHP version:**
7.4.25

**Teampass version:** 
3.0.0.10

LDAP Settings:
image

Below is the sniffing result that occurs during LDAP synchronization:
image

terwarf:
Use user provided filter if configured in the LDAP-settings page instead of the default filter.
Should fix #3040

### DIFF
--- a/sources/users.queries.php
+++ b/sources/users.queries.php
@@ -2449,13 +2449,17 @@ if (null !== $post_type) {
             $teampassRoles = array();
             $adUsedAttributes = array('dn', 'mail', 'givenname', 'samaccountname', 'sn', $SETTINGS['ldap_user_attribute'], 'memberof', 'name', 'displayname', 'cn', 'shadowexpire');
 
-            $users = $connection->query()->where([
-                ['objectclass', '=', 'top'],
-                ['objectclass', '=', 'person'],
-                ['objectclass', '=', 'organizationalperson'],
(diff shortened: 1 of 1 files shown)


hitenmandalia:
This is awesome.  Tested and confirmed it works.  Finally I can now look to test this and put into use.  Thank you :)

PR #3146, opened 2022-05-02
Change Requirements Readme

SlyRab:
Main.function.php contains union type of declaration. It has been avaible since php 8.0
I spend my whole day to understand why it's not working :(


nilsteampassnet:
Thank you.
I nevertheless decided to keep TP compliant with PHP7.4 until now

Issue #3163, opened 2022-05-16
[3.0.0.17] Generating keys - Encrypting keys - Personal items [0 - 100] - Stuck


reginfo:
### Steps to reproduce
1.Login as admin
2.Add new AD User

### Expected behaviour
Complete procedure

### Actual behaviour
Stuck on Generating Keys - Encrypting keys - Personal items [0 - 100]

<img width="609" alt="image" src="https://user-images.githubusercontent.com/4628571/168582069-3bcfa4c1-84fd-4242-b424-1b430ace8c8e.png">

### Server configuration
**Operating system**:
Ubuntu 20.04
**Web server:**
Apache/2.4.41 (Ubuntu)
**Database:**
mysql  Ver 15.1 Distrib 10.7.3-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
**PHP version:**
PHP 7.4.3
**Teampass version:** 
version 3.0.0.17
**Teampass configuration file:**
PLEASE attach to this issue the file `/includes/config/tp.config.php`.
[tp (1).txt](https://github.com/nilsteampassnet/TeamPass/files/8699591/tp.1.txt)

**Updated from an older Teampass or fresh install:**
Fresh

### Client configuration
**Browser:**
Chrome, Edge
**Operating system:**
Windows 11
#### Web server error log
```
Nothing in error.log

```

#### Log from the web-browser developer console (CTRL + SHIFT + i)
```
DevTools failed to load source map: Could not load content for http://192.168.xx.xx/plugins/summernote/summernote-bs4.css.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE

```

lollko:
Hi,

I have same issue 

when i check error log nothing happen , but if i check php. log i find somthing 

 PHP Fatal error:  Uncaught TypeError: base64_encode(): Argument #1 ($string) must be of type string, bool given in /var/www/TeamPass1/sources/main.functions.php:2738
Stack trace:
#0 /var/www/TeamPass1/sources/main.functions.php(2738): base64_encode()
#1 /var/www/TeamPass1/sources/main.queries.php(1893): decryptUserObjectKey()
#2 /var/www/TeamPass1/sources/main.queries.php(1747): continueReEncryptingUserSharekeysStep1()
#3 /var/www/TeamPass1/sources/main.queries.php(395): continueReEncryptingUserSharekeys()
[...]

PR #3172, opened 2022-05-19
3.0.0.17

dsnt02518:
Typo fix - resolves https://github.com/nilsteampassnet/TeamPass/issues/3163


dsnt02518:
This is not the only issue here. ...modal('hide') is definitely successfully called now, but isn't responding at the time it is called.

PR #3209, opened 2022-06-08
Fixing broken docker clone

jared-bloomer:
Docker clone commands fail because `/var/www/html` is not an empty directory. I have replaced the `docker clone` command with another way that will pull the code into the directory successfully.


nilsteampassnet:
Thank you

PR #3263, opened 2022-07-22
fix: cache tree table was lacking primary key

lucasapereira:


nilsteampassnet:
Thank you but already did this in a previous commit.
I didn't saw your PR

PR #3318, opened 2022-09-13
Translated most frequently used polish phrases

piechota1989:
Still a lot of things need to be translated


nilsteampassnet:
@piechota1989 Thank you for this.
Would it be possible to use https://poeditor.com/projects/view?id=433631?
I have now to inject manually those translations into POEditor ;(

Knoppix1:
> @piechota1989 Thank you for this. Would it be possible to use https://poeditor.com/projects/view?id=433631?

404 Error : https://poeditor.com/projects/view?id=433631 
I would like translate to french

PR #3418, opened 2022-11-17
Update Dockerfile

roccodelux:
Updated Docker version to support php 7.4.5 - richarvey/nginx-php-fpm:1.9.1

Added directive alpine-keys --allow-untrusted to add repo http://dl-3.alpinelinux.org/alpine/edge/testing.


nilsteampassnet:
Thank you

Issue #3267, opened 2022-07-28
CSRFP/LOG repository not create with installation


lfmgmsolutions:
### Steps to reproduce
1.Clone project 
2. Execute https://teampassUrl/install/install.php
3.Clic on start

### Expected behaviour
Directory "/includes/libraries/csrfp/log/" is writable == OK

### Actual behaviour
Directory "/includes/libraries/csrfp/log/" is writable  Path /var/www/html/teampass/includes/libraries/csrfp/log/ is not writable!

### Server configuration
DEBIAN 10

**Web server:**
APACHE2

**Database:**
MARIADB

**PHP version:**
8.2

**Teampass version:** 
3.0.0.18

**Teampass configuration file:**

**Updated from an older Teampass or fresh install:**
PLEASE attach to this issue the file `/includes/config/tp.config.php`.

### Client configuration
**Browser:**
Chrome

**Operating system:**
WIN11

#### Web server error log
```
Insert your webserver log here
```

#### Log from the web-browser developer console (CTRL + SHIFT + i)
```
Insert the log here and especially the answer of the query that failed.
```
![image](https://user-images.githubusercontent.com/97946975/181610502-423e93f1-3047-4743-97fa-dea1c34563c7.png)

you can add a log repository in your master please ?

![image](https://user-images.githubusercontent.com/97946975/181611746-2437cdbf-7166-47ae-9476-5c172eb219bd.png)

lfmgmsolutions:
@nilsteampassnet  i send a request for add a LOG directory pending install.

and I also see that you have LDAP problems I am a French Active Directory expert who offers you his help if necessary

Bye

karkazzul:
I have the same problem with the latest version:
![image](https://user-images.githubusercontent.com/23294229/185069405-b7e33e22-7edf-4ce9-8f8b-755aaa50b936.png)

mvillodres:
I have the same issue

IKarasev:
Yep, dumped myself to this issue as well. Not big of a problem, but annoying still.

ssavva05:
If you are using docker, to bypass the message it is required to create the folder in the docker container: 
sudo docker exec -it teampass-teampass-web-1 bash
mkdir -p /var/www/html/includes/libraries/csrfp/log/
chown -R nginx:nginx /var/www/html/includes/libraries/csrfp/log/

drbogar:
[...]

PR #3455, opened 2022-12-22
base image update

dormancygrace:
Update to use maintained base image (**latest** tag). It comes with php 8.1. 
Also fixed https://github.com/nilsteampassnet/TeamPass/issues/3267
DB image changed to smaller one (alpine-based)


nilsteampassnet:
Thank you

PR #3458, opened 2022-12-28
Make password not appear in plain text in ldap page

MasterWigu:
This modification makes avoids having the password displayed in plain text when testing the ldap configuration in the ldap page.

It simply changes the input field type from text to password.


nilsteampassnet:
Thank you

PR #3488, opened 2023-01-16
Export Folder name when exporting to CSV

eduardomozart:
I'm migrating from Teampass to sysPass but the CSV export from Teampass do not include the folder name, so it's not possible to identify the Client (concept equivalent to a "Folder" at Teampass) through the CSV file.


nilsteampassnet:
Thank you

PR #3511, opened 2023-02-05
fix bug russian.php

litded:
', с новой строки


nilsteampassnet:
Thank you but please use https://poeditor.com/projects/view?id=433631 for translation strings.
Merged in POEditor.

Issue #3448, opened 2022-12-13
Search item. Edit item from search. Unable to save (jQuery exception), maybe a bug in all v3? 


amilhub:
### Steps to reproduce
1. Search for an item from password main page.
2. Select an item, and then Edit.
3. Make some modifications and click Save. Impossible to save item.

### Expected behaviour
The item should be saved normally.

### Actual behaviour
When selecting an item comming from search, and edit with changes, unable to save.

jquery error: 
**jQuery.Deferred exception: itemsList[store.get(...).id] is undefined showItemEditForm**

Item is not being saved when comming from search. If i select the same item from the folder tree manually, it works as expected, I can save item normally.

Tested on 3.0.0.21 but also with latest master commit (and other older v3 versions), and issue is the same.

@nilsteampassnet please can you try to reproduce to check if this is a bug? 

### Server configuration
**Operating system**: CentOS 7.x

**Web server:** Apache 2.4

**Database:** MariaDB 10.2

**PHP version:** 7.4.32
**Teampass version:** 3.0.0.21

**Teampass configuration file:**

**Updated from an older Teampass or fresh install:**
Updated from older version.

### Client configuration
**Browser:** Any

**Operating system:** Any

#### Web server error log
```
Insert your webserver log here
```

#### Log from the web-browser developer console (CTRL + SHIFT + i)
```
Object { error: "", history: (41) […] }
index.php:8043:41
SHOW EDIT ITEM index.php:3894:51
Get privilege for folder false index.php:8390:47
>>> .item-details-card -- index.php:4091:48
SHOW EDIT ITEM false index.php:6128:47
jQuery.Deferred exception: itemsList[store.get(...).id] is undefined showItemEditForm@http://teampass_url.com/index.php?page=items:6176:49
@http://teampass_url.com/index.php?page=items:3926:33
e@http://teampass_url.com/plugins/jquery/jquery.min.js:2:30038
Deferred/then/l/</t<@http://teampass_url.com/plugins/jquery/jquery.min.js:2:30340
setTimeout handler*Deferred/then/l/<@http://teampass_url.com/plugins/jquery/jquery.min.js:2:30549
c@http://teampass_url.com/plugins/jquery/jquery.min.js:2:28327
add@http://teampass_url.com/plugins/jquery/jquery.min.js:2:28605
Deferred/then/<@http://teampass_url.com/plugins/jquery/jquery.min.js:2:30645
[...]

PR #3524, opened 2023-02-14
fix endless "in progress" bar when editing item

kultur-burgenland:
Fix for bug when trying to edit an item and "In progress" bar will spin endlessly, providing the following error in console: jQuery.Deferred exception: Cannot read properties of undefined (reading 'rights') TypeError: Cannot read properties of undefined (reading 'rights') at showItemEditForm

Also possible fix for https://github.com/nilsteampassnet/TeamPass/issues/3448


nilsteampassnet:
Thank you

Issue #2470, opened 2018-11-07
Unsent mails, please help


mathzen:
Well I encounter exactly the same problem as in this thread : 
https://github.com/nilsteampassnet/TeamPass/issues/2171

But the solution isn't detailed, so here I am... Thanks in advance !

### Steps to reproduce
1. share and request mails are sent instantly and properly
2. other mail queued are sent after manually triggered

### Server configuration
**Operating system**: 
Windows 2012 R2

**Web server:** 
Apache 2.4.23

**Database:**
MySQL 5.7.14

**PHP version:**
7.0.10

**Teampass version:** 
2.1.26

**Teampass configuration file:**
<?php
    global $SETTINGS;
    $SETTINGS = array (
        'max_latest_items' => '10',
        'enable_favourites' => '1',
        'show_last_items' => '1',
        'enable_pf_feature' => '0',
    'log_connections' => '0',
        'log_accessed' => '1',
        'time_format' => 'H:i:s',
        'date_format' => 'd/m/Y',
        'duplicate_folder' => '0',
    'item_duplicate_in_same_folder' => '1',
    'duplicate_item' => '1',
    'number_of_used_pw' => '0',
        'manager_edit' => '1',
        'cpassman_dir' => 'C:/wamp/www/TeamPass',
        'cpassman_url' => 'http://localhost/TeamPass',
        'favicon' => 'http://localhost/TeamPass/favico.ico',
        'path_to_upload_folder' => 'C:/wamp/www/TeamPass/upload',
        'url_to_upload_folder' => 'http://localhost/TeamPass/upload',
        'path_to_files_folder' => 'C:/wamp/www/TeamPass/files',
        'url_to_files_folder' => 'http://localhost/TeamPass/files',
        'activate_expiration' => '0',
        'pw_life_duration' => '0',
        'maintenance_mode' => '0',
        'enable_sts' => '0',
        'encryptClientServer' => '1',
        'cpassman_version' => '2.1.26',
        'ldap_mode' => '0',
        'ldap_type' => 'windows',
        'ldap_suffix' => 'secret',
        'ldap_domain_dn' => 'secret',
        'ldap_domain_controler' => 'secret',
        'ldap_user_attribute' => '0',
        'ldap_ssl' => '0',
        'ldap_tls' => '0',
        'ldap_elusers' => '0',
        'ldap_search_base' => '0',
        'richtext' => '0',
        'allow_print' => '0',
        'roles_allowed_to_print' => '0',
[...]

Issue #2171, opened 2018-03-27
unsent emails never send until triggered manually


ggzengel:
mail config works:
1. share and request mails are send instantly and properly
2. other mail queued are send after manually triggered

Even after days they are not send.
I installed postfix locally for queuing but never see a request from apache until sending mail manually.

# Version

Version: 2.1.27.11

```
# dpkg-query --list | grep -e "php\|apache"
ii  apache2                         2.4.25-3+deb9u3                amd64        Apache HTTP Server
ii  apache2-bin                     2.4.25-3+deb9u3                amd64        Apache HTTP Server (modules and other binary files)
ii  apache2-data                    2.4.25-3+deb9u3                all          Apache HTTP Server (common files)
ii  apache2-utils                   2.4.25-3+deb9u3                amd64        Apache HTTP Server (utility programs for web servers)
ii  libapache2-mod-php7.0           7.0.27-0+deb9u1                amd64        server-side, HTML-embedded scripting language (Apache 2 module)
ii  php-common                      1:49                           all          Common files for PHP packages
ii  php7.0-bcmath                   7.0.27-0+deb9u1                amd64        Bcmath module for PHP
ii  php7.0-cli                      7.0.27-0+deb9u1                amd64        command-line interpreter for the PHP scripting language
ii  php7.0-common                   7.0.27-0+deb9u1                amd64        documentation, examples and common module for PHP
ii  php7.0-curl                     7.0.27-0+deb9u1                amd64        CURL module for PHP
ii  php7.0-gd                       7.0.27-0+deb9u1                amd64        GD module for PHP
ii  php7.0-json                     7.0.27-0+deb9u1                amd64        JSON module for PHP
ii  php7.0-mbstring                 7.0.27-0+deb9u1                amd64        MBSTRING module for PHP
ii  php7.0-mcrypt                   7.0.27-0+deb9u1                amd64        libmcrypt module for PHP
ii  php7.0-mysql                    7.0.27-0+deb9u1                amd64        MySQL module for PHP
ii  php7.0-opcache                  7.0.27-0+deb9u1                amd64        Zend OpCache module for PHP
ii  php7.0-readline                 7.0.27-0+deb9u1                amd64        readline module for PHP
ii  php7.0-xml                      7.0.27-0+deb9u1                amd64        DOM, SimpleXML, WDDX, XML, and XSL module for PHP
```

```
# uname -a
[...]

PR #3554, opened 2023-02-26
Seperated cron job and Email administrators upon any user login

JunkyDeveloper:
Because i want to send all backlog emails but not get informed everytime if a new user login into the system.
So I added at cron configuration a new lever which enables sending the mail backlog or not.
Like meantioned in #2470 and #2171

Only thing which is missing is the language file.
Would be nice if you can show me how add strings into the language files via the editor, because I think I will contribute more things in the future until it is perfect for our environment.


nilsteampassnet:
Merged, Thank you.

For translation, please use POEditor, see https://github.com/nilsteampassnet/TeamPass/discussions/3580

PR #3565, opened 2023-03-03
fixed hard-coded table prefix and backticks in strings

jvrodrigues:
Hello,

This is a very simple patch that may fix some issues when running the upgrade script of TeamPass.

On these queries the table prefix was hardcoded to "teampass_" instead of the variable $pre

Also, backticks were replaced by single quotes on the strings.
I was getting a php error: "Unknown column 'bulgarian' in 'where clause'"


nilsteampassnet:
Thank you.

PR #3577, opened 2023-03-08
fixed missing .php files for importing keepass xml export

11iq:


nilsteampassnet:
Thank you.
Fixed by importing new dependencies

PR #3579, opened 2023-03-10
Another fix for endless "in progress" bar when editing item

kultur-burgenland:
Another fix for bug when trying to edit an item and "In progress" bar will spin endlessly, providing the following error in console: jQuery.Deferred exception: Cannot read properties of undefined (reading 'rights') TypeError: Cannot read properties of undefined (reading 'rights') at showItemEditForm

Also possible fix for https://github.com/nilsteampassnet/TeamPass/issues/3448

Bildschirm­foto 2023-03-10 um 11 51 54

Bildschirm­foto 2023-03-10 um 11 52 05


nilsteampassnet:
Thank you

PR #3591, opened 2023-03-16
Fix disabling admin role

nfsprodriver:
it checked enabling action so it was impossible to enable an admin, if there was exactly one (with email and pw)


nilsteampassnet:
Thank you

PR #3598, opened 2023-03-20
Do not write ad groups into fonction id and

nfsprodriver:
... also remove all AD groups if none is assigned for user


nilsteampassnet:
Thank you

PR #3635, opened 2023-04-03
Do not store ad groups in edit screen

nfsprodriver:
Keep them in the roles_from_ad_groups field only
Applies to https://github.com/nilsteampassnet/TeamPass/pull/3598


nilsteampassnet:
Thank you

PR #3637, opened 2023-04-03
Add possibility to define permission via roles

nfsprodriver:
like admin, manager etc.
Not in gui, needs to be added to the tp.config.php file manually. The following needles are searched inside the Role titles:
- admin_needle
- manager_needle
- tp_manager_needle
- read_only_needle

Additionally you can use 'exclude_user' as needle for usernames, which not should be touched.

### DIFF
--- a/sources/identify.php
+++ b/sources/identify.php
@@ -582,6 +582,11 @@ function identifyUser(string $sentData, array $SETTINGS): bool
                 WHERE id IN %li',
                 $superGlobal->get('user_roles', 'SESSION')
             );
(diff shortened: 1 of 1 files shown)


nfsprodriver:
Due to missing UI and DB parts (yet), we have the option to manually add the needles to the tp.config.php file. Whenever a role title contains a needle (e.g. admin_needle is set to "Admin"), a user beeing assigned to that role is set to an administrator during login. Same for the other roles, but only one can be applied (last role id in DB). If an admin role is removed (e.g. via AD), the user is no admin anymore. So you can control all privileges centralized.

PR #3648, opened 2023-04-12
Update installation.md

zolikonta:
Fixed PHP extension name, removed duplicate mysql extension, added missing curl extension.


nilsteampassnet:
Thank You

PR #3716, opened 2023-05-12
Container Crond Fix

### LINKED ISSUE #3715: Cron does not work in container as curently configured
### Steps to reproduce
1. pull repo
2. docker build -t tag repo
3. deploy container, install & configure
4. 

image

```bash
# This can be fixed post-install by running

apk add --update busybox-suid

echo '* * * * * php /var/www/html/sources/scheduler.php #Teampass scheduler' > /var/spool/cron/crontabs/nginx

# However ...
```
5. Cron service is not set up properly in the container by the build process, so it doesn't seem to work anyway.

### Expected behaviour
Cron and the cron job should be set up properly by the build process

Ideally, we should also be able to call http(s)://teampass.example.com/sources/scheduler.php by an external cron (having some access restrictions on it would be extra nice)

### Actual behaviour
Cron job is not set up properly.
Even after fixing, Cron service is not set up properly.

### Server configuration
(container build using the repo Dockerfile)
Alpine Linux v3.17

PHP version: 8.2.5
Memory limit: 128M
Memory usage: 575.33 KB
Maximum time execution: 120
Maximum file size upload: 100M

Server version: 10.11.2-MariaDB-1:10.11.2+maria~ubu2204
Database size: 0.92MB

**Teampass version:** 
3.0.7

Andrei-Paul:
Added crontab suid fix.
Added missing crontab file fix.
Added supervisord crond configuration.
Fixes #3715

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,13 @@ RUN echo && \
   echo "max_execution_time = 120" >> /usr/local/etc/php/conf.d/docker-vars.ini && \
 echo
 
+RUN echo && \
+  # Fix 'Error when trying to read crontab : crontab: must be suid to work properly'
+  apk add --no-cache --update busybox-suid && \
+  # Fix 'Error when trying to read crontab : crontab: can't open 'nginx': No such file or directory'
+  echo '* * * * * php /var/www/html/sources/scheduler.php #Teampass scheduler' > /var/spool/cron/crontabs/nginx && \
+echo
+
 # Fix API URL, BUG: API not working in container. #2100
 # Search last } and insert configuration rows before
 RUN sed -i "/^}/i \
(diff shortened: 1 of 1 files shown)


Andrei-Paul:
*Sigh*... Does a "verbal" sign-off count? I confirm that this is, indeed, my commit.

dormancygrace (changes requested):
Please, just add `busybox-suid` to tenth line of Dockerfile

nilsteampassnet:
@Andrei-Paul THank you for your PR.
Is it currently working?

Andrei-Paul:
It is working (my current instance is using the container built with this amended Dockerfile ), but the cron job is throwing a lot of "deprecated" warnings in the container logs, show in https://github.com/nilsteampassnet/TeamPass/issues/3715#issuecomment-1545802405

Either those should be fixed, or the output of the cron job should be silenced so debugging other issues won't be made really hard because of the noise.

PR #3789, opened 2023-07-12
Fix problem with One-Time-View

rsalas:
The redirection problem is solved when we are sending a One-Time-View link


nilsteampassnet:
Thank you

PR #3790, opened 2023-07-13
Solve spanish mail password

rsalas:
Solve problem the email send in spanish with #contraseña# instead of #password# and other translations


nilsteampassnet:
Hello
Thank you for your contribution.
Nevertheless all translation strings are managed from poeditor website.
Please contribute from this place.
Explanation is here : https://github.com/nilsteampassnet/TeamPass/discussions/3580
Thanks

PR #3829, opened 2023-08-16
Changing misleading text when logging/registering as an ldap user

MohanadAbugharbia:
Hello,

I'm making these changes because of direct feedback from our users who were confused as to why they couldn't login with their Active Directory credentials.

First time contributing.
Please let me know if I need to take any more steps


MohanadAbugharbia:
Im so sorry, I guess I missed that.

Thank you for making these changes anyway ☺️

PR #3844, opened 2023-08-31
Removed base64_decode used against plaintext in downloadFile.php

tomamoto:
This was resulting in either garbled filenames when trying to download attachments, or causing the server to outright refuse the download due URL-unfriendly characters.  It seems self-evident by the use of the basename() that the 'name' and 'extension' fields are stored in plaintext in the {prefix}_files table, so I can only assume this was an oversight and that it's still in the code since 3.0.0 because few people use the attachments feature.

### DIFF
--- a/sources/downloadFile.php
+++ b/sources/downloadFile.php
@@ -97,7 +97,7 @@ if (isset($_GET['pathIsFiles']) && (int) $get_pathIsFiles === 1) {
         decryptUserObjectKey($file_info['share_key'], $_SESSION['user']['private_key'])
     );
     // Set the filename of the download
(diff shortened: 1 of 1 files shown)


tomamoto:
@nilsteampassnet Is there some misunderstanding here?  I sent this pull request a while back after it helped me fix an issue where my users were unable to download their attachments.  I am able to accept being told I'm wrong or that my solution is inadequate.  I would just appreciate some feedback at this point, even if the merge itself might now be outdated.

nilsteampassnet:
Sorry @tomamoto for coming back to you so late.
Thank you for your PR and indeed it should have fixed the issue of several users.
In between another way to sanitize the data has been implemented.
Regards

PR #3887, opened 2023-09-28
Fix infinite up corner loading indicator firefox

alarido:
The up corner loading indicator keeps loading for a lot of time in firefox, with this fix it finish and behaves as expected


nilsteampassnet:
Thank you

Issue #3711, opened 2023-05-11
Ldap Roles Search uses 'Additional User DN' as base


Andrei-Paul:
### Steps to reproduce
1. Configure LDAP for OpenLDAP
2. Set Base DN = dc=example
3. Set Additional User DN = dc=people
4. LDAP group object filter = (objectClass=groupOfNames)
5. Go to Roles, Click LDAP Synchronization

### Expected behaviour
I should see roles from subtrees of dc=example that match the filters I set.

Ideally, there should be a "Set Additional Role DN" and that should be used as base for roles search.

### Actual behaviour
I see in the OpenLDAP server logs that the roles search used base dc=people,dc=example for the search

### Server configuration
(container build using the repo Dockerfile)
Alpine Linux v3.17

PHP version: 8.2.5
Memory limit: 128M
Memory usage: 575.33 KB
Maximum time execution: 120
Maximum file size upload: 100M

Server version: 10.11.2-MariaDB-1:10.11.2+maria~ubu2204
Database size: 0.92MB

**Teampass version:** 
3.0.7

**Teampass configuration file:**
[...]
    'ldap_mode' => '1',
    'ldap_type' => 'OpenLDAP',
    'ldap_suffix' => '0',
    'ldap_domain_dn' => '0',
    'ldap_domain_controler' => '0',
    'ldap_user_attribute' => 'uid',
    'ldap_ssl' => '1',
    'ldap_tls' => '0',
    'ldap_search_base' => '0',
    'ldap_port' => '636',
    'ldap_object_class' => '0',
    'ldap_new_user_is_administrated_by' => '0',
    'ldap_and_local_authentication' => '1',
    'ldap_user_dn_attribute' => 'dn',
    'ldap_dn_additional_user_dn' => 'dc=people',
    'ldap_user_object_filter' => '(objectClass=shadowAccount)(cn=*)(memberOf=cn=teampass,dc=services,dc=example)',
    'ldap_bdn' => 'dc=example',
    'ldap_hosts' => 'ldap.example',
    'ldap_password' => 'password.example',
    'ldap_username' => 'cn=teampass,dc=agents,dc=example',
    'ldap_tls_certifacte_check' => 'LDAP_OPT_X_TLS_NEVER',
    'enable_ad_users_with_ad_groups' => '1',
    'ldap_group_object_filter' => '(objectClass=groupOfNames)',
    'ldap_guid_attibute' => 'gidNumber',
[...]

#### From OpenLDAP
[...]

PR #3924, opened 2023-11-03
Remove 'additional_user_dn' when building the LDAP configuration array for role queries.

kcbieng:
The current code combines the 'Base DN' with the 'Additional User DN' settings when building the LDAP configuration array for role queries. This limits the search for Roles (groups) in the LDAP server to the OU or CN where the user objects are stored. However it is a common practice is to have groups and users in separate OUs or CNs. 

Therefore, I propose changing the logic in the LDAP configuration array to set the 'base_dn' in the array to the 'Base DN' user setting. 

In the future it would make sense to add an additional user setting called 'Additional Group DN' and junction that with the 'Base DN' user setting to enable group location filtering.


kcbieng:
#3711

kcbieng:
#3711

nilsteampassnet:
@kcbieng 
Thank you for the proposal :+1: 
I'm taking it into account as it sounds logic.

Issue #3964, opened 2023-11-28
Import XML from Keepass Don't Work


jucabalasilva:
### Steps to reproduce
1. Login in Teampass
2. Go to import
3. Choose Keepass
4.  Browser the file .xml

### Expected behaviour
Start import the password from XML generetad from KeePass

### Actual behaviour
Show the message: " Next error occurred
TypeError: Cannot read properties of null (reading 'length')

Raw answer from server:
null"

### Server configuration
**Operating system**: Oracle Linux 9.3

**Web server:** Apache 2.4.57

**Database:** MariaDB 10.5.22

**PHP version:** PHP 8.1.25

**Teampass version:** 3.1.0.6

**Teampass configuration file:**
```
<?php
global $SETTINGS;
$SETTINGS = array (
    'max_latest_items' => '10',
    'enable_favourites' => '1',
    'show_last_items' => '1',
    'enable_pf_feature' => '1',
    'log_connections' => '1',
    'log_accessed' => '1',
    'time_format' => 'H:i:s',
    'date_format' => 'd/m/Y',
    'duplicate_folder' => '0',
    'item_duplicate_in_same_folder' => '0',
    'duplicate_item' => '0',
    'number_of_used_pw' => '3',
    'manager_edit' => '1',
    'cpassman_dir' => '/var/www/TeamPass',
    'cpassman_url' => 'https://lnxteampassprd01.xxxxxx.com.br',
    'favicon' => 'https://lnxteampassprd01xxxxxx.com.br/favicon.ico',
    'path_to_upload_folder' => '/var/www/TeamPass/upload',
    'path_to_files_folder' => '/var/www/TeamPass/files',
    'url_to_files_folder' => 'http://lnxteampassprd01xxxxxxl.com.br/files',
    'activate_expiration' => '0',
    'pw_life_duration' => '0',
    'maintenance_mode' => '0',
    'enable_sts' => '1',
    'encryptClientServer' => '1',
    'teampass_version' => '3.1.0',
    'ldap_mode' => '1',
    'ldap_type' => 'ActiveDirectory',
    'ldap_suffix' => '0',
    'ldap_domain_dn' => '0',
    'ldap_domain_controler' => '0',
    'ldap_user_attribute' => 'samaccountname',
    'ldap_ssl' => '0',
    'ldap_tls' => '0',
    'ldap_search_base' => '0',
    'ldap_port' => '389',
    'richtext' => '0',
    'allow_print' => '0',
[...]

PR #3982, opened 2023-12-13
Fix Keepass Import Uploads

kcbieng:
Per Issue #3964 Keepass file imports were failing with a Null(length) error. In troubleshooting I found that "upload.files.php" was failing in the Keepass section when it checked for the "user_token". Looking at the post payload sent to upload.files.php from I could see that the token was not being sent to the uploader. 

I found the up.settings.multipart_params being set for Keepass file uploads being set at line 522 of import.js.php. These settings were being set as an array. I compared this to the working CSV upload section and found there that the settings were being set individually. I'm not smart enough to know why this wasn't working, but I switched the Keepass settings to be set individually and the upload began working correctly.


nilsteampassnet:
@kcbieng 
Thank you, this is correct.
I'm merging this.

PR #4048, opened 2024-02-07
Syslog Item logging fix

Deadbeastrs:
The items log is not correctly forwarded to syslog since the inline comparison executed, when true, skips the action, attribute and itemno. This is easily fixed by adding proper curly brackets to the statement.


nilsteampassnet:
@Deadbeastrs 

Thank you

PR #4062, opened 2024-02-19
Fix 2FA HTML code displayed on login page

manusfreedom:
Fix 2FA HTML code displayed.


nilsteampassnet:
@manusfreedom 
Thank you

PR #4137, opened 2024-04-12
fix: wget raw content when downloading docker-compose.yml

paciook:
Tried to wget the provided link but got the whole page instead of the file content. Had to get the raw link from Github.


nilsteampassnet:
Thank you

PR #4140, opened 2024-04-16
fix: CERT_NAME shouldn't contain extension .crt

paciook:
CERT_NAME shouldn't contain .crt extension. Check https://hub.docker.com/r/jwilder/nginx-proxy SNI section.


nilsteampassnet:
Thank you

PR #4143, opened 2024-04-16
Update PHPMailer.php

Hard-3d:
With this correction, Cyrillic letters come in a readable form.


nilsteampassnet:
Thank you.
But merge not done because it is modifying an external library.
I have update the charset inside the call.

PR #4185, opened 2024-05-14
Fix for broken JSON response when ldap group id is binary

simonpoess:
changed placeholder in ldap_groups_roles queries to reflect the non-integer value

My Problem was, when using Teampass with Active Directory, I didn't get the right amount of Roles through the LDAP Synchronization, with Int I got only 3 roles through the Synchronization, so we had the idea to change the int value to a md5 hash.
In my case it worked and I got all Roles through the Active Directory. 
My next problem was, that Teampass mapped some of the Active Directory Roles with some Teampass Roles I imported through the User Synchronization. When I tried to map a Role to another Role and hit Refresh, the change was reverted. After a bit of research I found a similar issue where you posted a fix, most of it was already implemented but a placeholder change in roles.queries.php on line 746 was missing.

Additionally I had to change the placeholder in identify.php on line 1419 from %I to %s because the Roles didn't update on login. Through the change the User got their roles assigned on login and get updated after every Login.


simonpoess:
Had to force push for signing

Issue #4224, opened 2024-06-20
Deleting item from popup button empty all folders


antoweb:
### Steps to reproduce
1. Select an item from list
2. Delete item clicking on trash button popup (see screenshot)
3. Select a folder from left side, on the right side appear message : "No items to display"

![image](https://github.com/nilsteampassnet/TeamPass/assets/32364871/8ec4e143-d05c-4281-b65a-6ffc06ea0530)

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

### Server configuration
**Operating system**:

**Web server:**

**Database:**
mysql  Ver 15.1 Distrib 10.11.6-MariaDB
**PHP version:**
8.2.13

**Teampass version:** 
3.1.2

**Teampass configuration file:**

**Updated from an older Teampass or fresh install:**
PLEASE attach to this issue the file `/includes/config/tp.config.php`.

### Client configuration
**Browser:**
All

**Operating system:**
Red Hat 8.10

#### Web server error log
```
Insert your webserver log here
```

#### Log from the web-browser developer console (CTRL + SHIFT + i)
```
Insert the log here and especially the answer of the query that failed.
```

corentin-soriano:
PR: #4228

antoweb:
Hello, so we have to waiting a new release?

corentin-soriano:
Sorry, I commented on the wrong issue.
Do you encounter the problem systematically, or only on items imported from a keepass/csv?

In manual created item :
![image](https://github.com/nilsteampassnet/TeamPass/assets/107032768/59300718-6e8c-4b12-a6a4-984575f14a30)

In imported item:
![image](https://github.com/nilsteampassnet/TeamPass/assets/107032768/915e23e8-0a1c-4cfa-9811-ea2fb852a774)

data-item-key seems not to be correct.

corentin-soriano:
I understood the problem, I made the pull request #4230 with details.

antoweb:
Imported items trought csv

nilsteampassnet:
Closed automatically due to inactivity (no response for over 1 year).

PR #4230, opened 2024-06-24
Correct issues in delete items in list

corentin-soriano:
Commit "Avoid duplicate items in list after a delete":
image

Commit "Delete by id instead of key to gain reliability": #4224 
- The item key can be duplicated in database.
- The imported items all have a key of -1.
- The item id is the primary key, so it can't be null or duplicate. It's more reliable.

image

### DIFF
--- a/pages/items.js.php
+++ b/pages/items.js.php
@@ -1459,8 +1459,6 @@ $var['hidden_asterisk'] = '<i class="fa-solid fa-asterisk mr-2"></i><i class="fa
                     
                     // Refresh tree
                     refreshTree(folderId, true);
-                    // Load list of items
-                    ListerItems(folderId, '', 0);
                     // Close
                     if (closeItemCard === true) {
                         closeItemDetailsCard();
--- a/pages/items.js.php
+++ b/pages/items.js.php
@@ -2344,15 +2344,15 @@ $var['hidden_asterisk'] = '<i class="fa-solid fa-asterisk mr-2"></i><i class="fa
             event.preventDefault();
             // Delete item
             if (debugJavascript === true) {
-                console.info('SHOW DELETE ITEM '+$(this).data('item-key'));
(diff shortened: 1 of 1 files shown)


FengZh61:
I also encountered this problem in version 3.1.2.45, when I imported the data through the csv file, the value of item_key was -1, and I also found that the value of created_at was null, I solved the problem of the value of item_key being -1 by modifying the incoming sql, but from the data structure of the datasheet, the value of item_key was not guaranteed unique, if item_key is duplicated, will I encounter problems in other scenarios?

The fix to avoid duplicate items in the list after deletion is great!

corentin-soriano:
@FengZh61 My bad, little mistake in https://github.com/nilsteampassnet/TeamPass/pull/4273/files#diff-1139f18c144af50ac2f79fa33a40e72d5209d055b98c04ea5351f7832e76cd9a

Can you test this fix? #4322

FengZh61:
> @FengZh61 My bad, little mistake in https://github.com/nilsteampassnet/TeamPass/pull/4273/files#diff-1139f18c144af50ac2f79fa33a40e72d5209d055b98c04ea5351f7832e76cd9a
> 
> Can you test this fix? #4322

I'm not very proficient with git, I'll post my modifications, feel free to discuss.

- sources/import.queries.php#L378 
Add the following fields
```php
 'item_key' => uniqidReal(50),
 'created_at' => time(),
```
The purpose of this part of the modification is to assign a value to the item_key so that it will not be -1, and to assign a value to the creation time, which is null by default.

- sources/items.queries.php#L2215
Add the following fields
```php
 $aSet = array();
 //$aSet['created_at'] = time();
       foreach ($originalRecord as $key => $value) {
             $aSet['created_at'] = time();
             $aSet['item_key'] = uniqidReal(50);
             if ($key === 'id_tree') {
```
This part of the modification is mainly to avoid the problem of duplicate item_key when copying the item.

PR #4257, opened 2024-07-16
Fix user personnal folder creation.

corentin-soriano:
Always display run job button for personnal folders:
image

Fix personnal folder script:
- PHP Fatal error:  Uncaught Error: Class "ConfigManager" not found in /var/www/teampass/scripts/task_maintenance_users_personal_folder.php:42
- Only create folders for users who have personal_folder = 1
- Remove unused code.


DobroFenix:
Not relevant for the new version. There is now another problem with writing information about a personal folder to the database, in principle. I have not yet understood how to solve the problem - the sql query is not executed.

corentin-soriano:
> Not relevant for the new version. There is now another problem with writing information about a personal folder to the database, in principle. I have not yet understood how to solve the problem - the sql query is not executed.

@DobroFenix can you describe the reproduction context?
You cloned this branch and the personal folders are still not created?
Is the scheduler configured to start automatically?
No other blocked tasks (visible in the tasks section in admin)?
image

DobroFenix:
Since I could not run teampass locally, because even during the installation process, install/install.queries.php rained down with various errors and refused to install it. This time I launched TeamPass not locally, but inside the docker container - https://hub.docker.com/r/valterseu/teampass
In the settings, I turned on personal folders, created a user, at the top he had a notification that something was being done. But the personal folder never appeared. There are no running tasks, all tasks are completed.


image
image

corentin-soriano:
The fix is ​​not present in the docker image since it has not yet been validated and merged by Nils.

DobroFenix:
Got that. Thank you.
At least so...
It's a pity TeamPass failed to run locally.
I hope he shows up and puts together a working version.

Issue #4212, opened 2024-06-04
Error in the scheduler due to data entered by a user


corentin-soriano:
### Steps to reproduce
1. Login as user
2. Generate new keys
3. Fill in the public and private key fields with custom text not in base64 format
4. background_tasks___items_handler.php will fail.

### Expected behaviour
User can't fill form with non base64 public key.

### Actual behaviour
![image](https://github.com/nilsteampassnet/TeamPass/assets/107032768/0eb10921-4b2d-42fa-8a3f-58ca817e4109)

User table content :
```sql
mysql> SELECT * FROM teampass_users WHERE LENGTH(public_key) < 20 \G
*************************** 1. row ***************************
                    id: 10000541
                 login: csoriano
                    pw: ******************************************
      groupes_visibles:
              derniers: NULL
             key_tempo: ************************
        last_pw_change: 1715946371
               last_pw: NULL
                 admin: 0
           fonction_id: 18
     groupes_interdits:
        last_connexion: 1717491911
          gestionnaire: 0
                 email: ************@************
            favourites: NULL
          latest_items:
       personal_folder: 1
              disabled: 0
       no_bad_attempts: 0
can_create_root_folder: 0
             read_only: 0
             timestamp: 1717492814
         user_language: french
                  name: Corentin
              lastname: Soriano
           session_end: 1717521911
 isAdministratedByRole: 0
                   psk: NULL
                    ga: NULL
     ga_temporary_code: none
                avatar: NULL
          avatar_thumb: NULL
        upgrade_needed: 0
      treeloadstrategy: full
  can_manage_all_users: 0
          usertimezone: Europe/Paris
      agses-usercardid:
         encrypted_psk:
               user_ip: 10.0.0.1
      user_ip_lastdate: 1717491915
       yubico_user_key: none
        yubico_user_id: none
            public_key: test@125
           private_key: hnXJRuy+wsYKqKi+xzTcnQ==
               special: generate-keys
             auth_type: local
    is_ready_for_usage: 1
          otp_provided: 1
  roles_from_ad_groups: NULL
    ongoing_process_id: NULL
           mfa_enabled: 0
            created_at: 1715946371
            updated_at: NULL
            deleted_at: NULL
    keys_recovery_time: 1715946798
                aes_iv: NULL
```
Launch background_tasks___items_handler.php with WIP = true :
```log
[...]

PR #4268, opened 2024-07-26
Sanitize import priv/pub keys field.

corentin-soriano:
Fix for issue #4212


corentin-soriano:
@nilsteampassnet why have you reverted this fix in https://github.com/nilsteampassnet/TeamPass/commit/dd2ce469cc36e941de9018c2780babea6cba5d97 ?
Is there a bug ?

nilsteampassnet:
@corentin-soriano 
I don't really understand.
It seems I have merged this one but did not pulled it back.

PR #4280, opened 2024-08-12
BUGFIX - Cast all fields on API createUserJWT function call.

corentin-soriano:
Fix this error when authenticating in API:
```json
{"error":"AuthModel::createUserJWT(): Argument #14 ($allowed_folders) must be of type string, null given, called in \/var\/www\/teampass\/api\/Model\/AuthModel.php on line 155 Something went wrong! Please contact support.2"}
```


nilsteampassnet:
Thank you

PR #4282, opened 2024-08-13
Trim values to avoid break json.

corentin-soriano:
Avoid this error (tab in start of json values):
image


nilsteampassnet:
Thank you

PR #4285, opened 2024-08-13
Avoid bypass maximum_session_expiration_time parameter.

corentin-soriano:
Actually, user can bypass maximum_session_expiration_time parameter on login page:
image

image

End of session timestamp: 7723553611 -> Saturday 1 October 2214 23:33:31 GMT


nilsteampassnet:
Thank you

PR #4286, opened 2024-08-13
Add css/js script version to avoid cache issues on upgrade.

corentin-soriano:
Example after 2.x to 3.x upgrade (error in js console and infinite loading):
image

image

Users should clear their caches or CTRL+F5 without this fix.


nilsteampassnet:
Thank you

PR #4293, opened 2024-08-23
Fix copy password log.

corentin-soriano:
Fix missing item id on copy password log.
Payload before my change:
```json
{"label":"test","user_id":"10000001","action":"at_password_copied","login":"csoriano"}
```
image

And after:
```json
{"id":"6","label":"test","user_id":"10000001","action":"at_password_copied","login":"csoriano"}
```


nilsteampassnet:
Thank you

PR #4273, opened 2024-08-01
UI - Improve items page UX to see tree/items list and items details simultaneously.

corentin-soriano:
Our users regret not being able to navigate between different items without a full screen popup like in teampass 2.x.
What I propose in this PR is to display the item to the right of the tree structure which is reduced only when an item is opened and returns to its initial size afterwards.

- No change on default item view:
image

- Open an item:
image

- After closing item, the default view is reverted.

- Open item details after edit or create.
image

- Same height for tree and items list containers. Scroll in items list/tree but not on global web page:
image

- Always refresh items list / item details after edition.

- Remove this ugly and confusing empty line (issue with buttons solved thanks to fixed height):
image

- Add optional variable to change page title:
image

- Correct multiple errors when deleting item from search.

- Mouse pointer on these links:
image

- Allow select text on URL field and add copy/open buttons:
image

- Remove useless buttons:
image

- Dark/light theme:
image
image

PR #4322, opened 2024-09-16
Delete by id instead of key to gain reliability.

corentin-soriano:
Fix small regression in https://github.com/nilsteampassnet/TeamPass/pull/4273/files#diff-1139f18c144af50ac2f79fa33a40e72d5209d055b98c04ea5351f7832e76cd9a


nilsteampassnet:
Thank you

PR #4329, opened 2024-09-16
Update oauth.php

kcbieng:
Mismatch in the constructor parameters when instantiating the AzureAuthController class.


nilsteampassnet:
Merged.
Thank you

PR #4332, opened 2024-09-17
Add support of generic OAuth2 provider in addition to MS Azure, auto-login and fix security issue.

corentin-soriano:
I am able to login with my account with a standard oauth2 provider after updating my account auth type (need to restore my keys after login):
```sql
UPDATE users SET auth_type = 'oauth2' WHERE login = 'corentin';
```

With the second commit, we can enable sso auto login. Bypass autologin with URL index.php?page=items&loginForm or logout button.

This "password" seems to be insecure but I don't know if we can do anything really better:
```php
            // manage cryto ID
            function hashUserId(userId) {
                const hash = CryptoJS.SHA256(userId);
                return hash.toString(CryptoJS.enc.Hex).substring(0, 16);
            }
```

There is still a step to migrate from a password (ldap or local) account to an SSO account.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -108,7 +108,7 @@
         },
         {
             "type": "path",
-            "url": "includes/libraries/teampassclasses/azureauthcontroller",
+            "url": "includes/libraries/teampassclasses/oauth2controller",
             "options": {
                 "symlink": false
             }
@@ -146,7 +146,7 @@
         "teampassclasses/language": "dev-master",
         "teampassclasses/ldapextra": "dev-master",
         "teampassclasses/passwordmanager": "dev-master",
-        "teampassclasses/azureauthcontroller": "dev-master",
+        "teampassclasses/oauth2controller": "dev-master",
         "teampassclasses/configmanager": "dev-master",
         "teampassclasses/emailservice": "dev-master",
         "illuminate/contracts": "^10.32",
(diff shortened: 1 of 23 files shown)


nilsteampassnet:
> This "password" seems to be insecure but I don't know if we can do anything really better:
Yes but it is meaningless as it is not possible to get logged in with an oauth2 account using the normal login credentials.
So let's say risk is really mitigated

> There is still a step to migrate from a password (ldap or local) account to an SSO account.
Yes it is in the todo list.
I'm waiting for usage feedback for a couple of weeks and if ok then I will start this.

corentin-soriano:
> > This "password" seems to be insecure but I don't know if we can do anything really better:
>
> Yes but it is meaningless as it is not possible to get logged in with an oauth2 account using the normal login credentials.
> So let's say risk is really mitigated
> I didn't found any better solution until now...

The only limitation is DB encryption.
However I think it's still better than having the user's AD password in the browser's local storage and in the PHP session and this system removes the complexity of managing private keys since they will never change again.

> > There is still a step to migrate from a password (ldap or local) account to an SSO account.
>
> Yes it is in the todo list.
> I'm waiting for usage feedback for a couple of weeks and if ok then I will start this.

I think it can easily be implemented on the same basis as LDAP password synchronization. I'll try to take a few moments to test.

corentin-soriano:
With the last commit, on first oauth2 login, this form will appear:
image

And password are passwords are readable!
image

PR #4334, opened 2024-09-17
Fix missing folder icons.

corentin-soriano:
Fix for folders created without icon:
image


nilsteampassnet (approved):

nilsteampassnet:
Yes correct.
I've been too fast here.
Thank you, my tests were not covering all cases.

PR #4339, opened 2024-09-18
Auto select jstree folder on path elems click.

corentin-soriano:
Auto select folder in the jstree:
image

Replace obsolete onclick by event listener.

Select current folder instead of parent folder on the create folder form:
image
image


nilsteampassnet (approved):

corentin-soriano:
The last commit fix the selected folder on the edit form when item is opened with direct link:
image

Edit item:
image

The first visible folder is auto selected instead of current selected folder:
image

nilsteampassnet (approved):

PR #4359, opened 2024-09-25
Vulnerability log injection

nilsteampassnet:
It has been reported a vulnerability in items.log.php

> Here, it is important to note that in the payload sent to the server, it is speci昀椀ed that the user with ID 1
(here the admin account) would have created a folder. In the following example, the operation is
carried out with a simple user, using a script to supply the payload in clear text.
python3 teampass_poc.py request --uri "/sources/items.logs.php" --data '{"id": 1,"label":
"testtest","user_id": "1","action": "at_creation","login": "admin"}' --key "<encryption key
or cookie>" --type "log_action_on_item" <target>

The purpose of the PR is to reduce the context of usage of this log feature to only its use case (log quick icon password copy)

### DIFF
--- a/includes/config/include.php
+++ b/includes/config/include.php
@@ -28,7 +28,7 @@
 
 define('TP_VERSION', '3.1.2');
 define("UPGRADE_MIN_DATE", "1727110744");
-define('TP_VERSION_MINOR', '112');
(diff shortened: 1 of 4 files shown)


corentin-soriano (approved):
Looks good to me!

PR #4371, opened 2024-09-26
Fixing vulnerability with clear password in email log

nilsteampassnet:
Fixing vulnerability with clear password in email log

> When an account is created on the application, a sixteen-character password is generated and sent in
cleartext by e-mail to the address used to generate the account.
These actions are logged in the database (in the teampass_background_tasks table), along with the
contents of the email containing the plaintext password.

### DIFF
--- a/sources/main.functions.php
+++ b/sources/main.functions.php
@@ -1103,7 +1103,7 @@ function getStatisticsData(array $SETTINGS): array
  * @param string $body          email message
  * @param string $email         email
  * @param string $receiverName  Receiver name
- * @param array  $SETTINGS      settings
+ * @param string $encryptedUserPassword      encryptedUserPassword
  *
  * @return void
  */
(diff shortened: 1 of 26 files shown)


sources/main.queries.php
```diff
@@ -431,6 +431,7 @@ function mailHandler(string $post_type, /*php8 array|null|string */$dataReceived
         * Send email
         */
         case 'mail_me'://action_mail
+            error_log('mail_me');
```
corentin-soriano, line 434: This debug log needs to be removed.

sources/main.functions.php
```diff
@@ -1103,17 +1103,19 @@ function getStatisticsData(array $SETTINGS): array
...
 function prepareSendingEmail(
     $subject,
     $body,
     $email,
-    $receiverName = ''
+    $receiverName = '',
+    $encryptedUserPassword = ''
 ): void 
 {
+    error_log('prepareSendingEmail');
```
corentin-soriano, line 1118: This debug log needs to be removed.

corentin-soriano (changes requested):

corentin-soriano (approved):

PR #4373, opened 2024-09-27
Vulnerability using API authorization with SQL Injection

nilsteampassnet:
Fixing vulnerability regarding API authorize process

> A SQL injection has been detected on the API. This injection occurs on the endpoint used to
authenticate users on the API ( /api/index.php/authorize ). It turns out that at code level, two
vulnerable queries are used to authenticate users. These SQL queries are performed in the
getUserAuth function of the /api/Model/AuthModel.php 昀椀le.
As it can be seen, a concatenation is performed in the PHP code with one of the parameters ( login )
controlled by the user. However, it remains di昀케cult to exploit, as the anti XSS 昀椀lters applied to user
input do not allow the request to be manipulated in order to add code later.
However, a second request may allow controlled injection.

The fix consists in removing the old condition of auth with a generic key, which is not allowed any more.
THis part of code has been removed and replaced by an error code.
In front, the API generic keys page has already a warning showing they should not be used any more.


api/Model/AuthModel.php
```diff
@@ -68,53 +68,12 @@ public function getUserAuth(string $login, string $password, string $apikey): ar
         if (empty($inputData['login']) === true || empty($inputData['apikey']) === true) {
             return ["error" => "Login failed.", "info" => "Empty entry"];
         }
-        
+
         // Check apikey
         if (empty($inputData['password']) === true) {
```
corentin-soriano, line 73: Do you think we could merge the two if like this:
```php
        if (empty($inputData['login']) === true || empty($inputData['apikey']) === true || empty($inputData['password']) === true) {```

**Review by corentin-soriano — APPROVED** [review]
I approve on this scope if you want to merge it however there are other SQL queries nested with PHP which are potentially vulnerable.

**Comment by nilsteampassnet** [comment]
> I approve on this scope if you want to merge it however there are other SQL queries nested with PHP which are potentially vulnerable.

Any example?

**Comment by corentin-soriano** [comment]
> > I approve on this scope if you want to merge it however there are other SQL queries nested with PHP which are potentially vulnerable.
> 
> 
> 
> Any example?
> 
> 

Line 81

**Comment by nilsteampassnet** [comment]
Yes I see what you mean.
Nevertheless this variable is sanitized before and no vulnerability exists on this request.
I have tested with postman and sql injection is correctly escaped. 
I'm thinking about adding an escape in the dB class just in case.

File: api/Model/AuthModel.php [hunk 2]
```diff
@@ -266,23 +272,32 @@ private function buildUserFoldersList(array $userInfo): array
                 }
             }
         }
-        
+        //error_log(print_r($userFunctionId, true));
```
corentin-soriano, line 275: Can you remove this debug line ?

corentin-soriano (approved):
Great work, thank you!

api/Model/AuthModel.php
nilsteampassnet: Done

PR #4377, opened 2024-09-28
Profile fixes

evertton:
- **Fixed update user language, when underline is a character in language selector**
  - When the language "Portuguese (Brazil)" was selected, the prevent LFI regular expression removed the underline character, so whenever the user chose this language they ended up receiving English as a fallback.

    ```php
    echo $inputData['language']; // portuguese_br
    
    // Prevent LFI.
    $inputData['language'] = preg_replace('/[^a-z]/', "", $inputData['language']);
    
    echo $inputData['language']; // portuguesebr
    ```

- **Fixes timezone selection display in profile**
  - The way the logic was before, the system timezone was always selected. When selecting "America/Maceio" timezone in the profile and saving, the next time it was displayed, "UTC" appeared again. It was necessary to choose the preferred timezone whenever making any edit.

### DIFF
--- a/pages/profile.php
+++ b/pages/profile.php
@@ -399,15 +399,16 @@ foreach ($session->get('user-roles_array') as $role) {
                                         <label class="col-sm-10 control-label"><?php echo $lang->get('timezone_selection');?></label>
                                         <div class="col-sm-10">
                                             <select class="form-control" id="profile-user-timezone">
-                                                <?php
-                                                foreach ($zones as $key => $zone) {
-                                                    echo '
-                                                <option value="' . $key . '"',
-                                                    $session->has('user-timezone') && $session->get('user-timezone') === $key ? ' selected' :
-                                                    (($session->has('user-timezone') && $session->get('user-timezone') === 'not_defined') && isset($SETTINGS['timezone']) && $SETTINGS['timezone'] === $key ? ' selected' : ''),
-                                                '>' . $zone . '</option>';
-                                                }
-                                                ?>
+                                                <?php foreach ($zones as $key => $zone): ?>
+                                                    <option value="<?php echo $key; ?>"<?php 
+                                                        if ($session->has('user-timezone'))
+                                                            if($session->get('user-timezone') === $key)
+                                                                echo ' selected';
+                                                            elseif ($session->get('user-timezone') === 'not_defined')
+                                                                if (isset($SETTINGS['timezone']) && $SETTINGS['timezone'] === $key)
+                                                                    echo ' selected';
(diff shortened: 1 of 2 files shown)


pages/profile.php
```diff
@@ -396,16 +396,15 @@
...
                                             <select class="form-control" id="profile-user-timezone">
                                                 <?php
                                                 foreach ($zones as $key => $zone) {
                                                     echo '
                                                 <option value="' . $key . '"',
-                                                    $session->has('user-timezone') && $session->get('user-timezone') && null !== $session->get('user-timezone') && $session->get('user-timezone') === $key ?
-                                                    ' selected' :
-                                                    (isset($SETTINGS['timezone']) === true && $SETTINGS['timezone'] === $key ? ' selected' : ''),
+                                                    $session->has('user-timezone') && $session->get('user-timezone') === $key ? ' selected' :
+                                                    (($session->has('user-timezone') && $session->get('user-timezone') === 'not_defined') && isset($SETTINGS['timezone']) && $SETTINGS['timezone'] === $key ? ' selected' : ''),
```
corentin-soriano, line 407: It should work but I find that this ternary condition is complicated to read due to its length.

corentin-soriano (approved):

pages/profile.php
nilsteampassnet: Can't this be replaced by

```
$userTimezone = $session->get('user-timezone');
$selected = ($userTimezone === $key) ? ' selected' : '';

if ($userTimezone === 'not_defined' && isset($SETTINGS['timezone']) && $SETTINGS['timezone'] === $key) {
    $selected = ' selected';
}

return $selected;

```

evertton: I agree with you, is complicated to read. I would have liked to write the code like below, but I chose to stick with the standard adopted earlier.

```php
<?php foreach ($zones as $key => $zone): ?>
    <option value="<?php echo $key; ?>"<?php 
    if ($session->has('user-timezone'))
        if($session->get('user-timezone') === $key)
            echo ' selected';
        elseif ($session->get('user-timezone') === 'not_defined')
            if (isset($SETTINGS['timezone']) && $SETTINGS['timezone'] === $key)
                echo ' selected';
    ?>><?php echo $zone; ?></option>
<?php endforeach; ?>
```

Should I commit the rewrite?

nilsteampassnet: Yes seems nice.

Can you please also sign your commit (see https://docs.github.com/fr/authentication/managing-commit-signature-verification/about-commit-signature-verification)

Issue #4316, opened 2024-09-12
Error in the character encoding in “User Definition”.


gonzalodeniz:
### Steps to reproduce
1. Login as admin
2. Open Users window
3. Edit a user
4. Modify the first and last name with the following characters: **áéíóú**
5. Edit the user again
6. Now the first and last name look like this: **&aacute  ;&eacute ;&iacute ; &oacute ; &uacute ;**

<img width="482" alt="image" src="https://github.com/user-attachments/assets/3c7fa0a9-d11c-4bb2-a631-d6869e5f5bff">

### Expected behaviour
When editing, the characters must appear correctly encoded.
<img width="443" alt="image" src="https://github.com/user-attachments/assets/3afb3f2b-d53b-4e1b-ac54-22a6dd3fd96f">

### Actual behaviour
Characters are not displayed correctly encoded
<img width="611" alt="image" src="https://github.com/user-attachments/assets/b87655bd-cf81-415f-a7c0-da7617110037">

### Server configuration
**Operating system**:Ubuntu 22.04

**Web server:**Apache 2.4

**Database:**MariaDB 10

**PHP version:**8.2

**Teampass version:** 3.1.2.71

#### Web server error log
```
None
```

#### Log from the web-browser developer console (CTRL + SHIFT + i)
```
None
```

gonzalodeniz:
Fix:

sources/users.queries.php
Replace lines 1360 y 1361:
```
$arrData['name'] = empty($rowUser['name']) === false && $rowUser['name'] !== NULL ? htmlspecialchars_decode($rowUser['name'], ENT_QUOTES) : '';
                $arrData['lastname'] = empty($rowUser['lastname']) === false && $rowUser['lastname'] !== NULL ? htmlspecialchars_decode($rowUser['lastname'], ENT_QUOTES) : '';
```                
by

```
$arrData['name'] = empty($rowUser['name']) === false && $rowUser['name'] !== NULL ? html_entity_decode($rowUser['name'], ENT_QUOTES, 'UTF-8') : '';
                $arrData['lastname'] = empty($rowUser['lastname']) === false && $rowUser['lastname'] !== NULL ? html_entity_decode($rowUser['lastname'], ENT_QUOTES, 'UTF-8') : '';
```

I have tried to upload a branch with the solution in order to request a pull requests but I don't have permission.

```
git push origin fix/4316-encoding-issue
[...]

PR #4383, opened 2024-09-30
Fix users encoding #4316

corentin-soriano:
- Safely remove html encoding by using virtual div element (data strings are always html safe):
```js
$('#form-login').val($('<div>').html(data.login).text());
```

- Fix issue where $antiXss->xss_clean resolve html encodings:
image

- Remove insecure htmlspecialchars_decode.

All fields seems to be secure:
image


nilsteampassnet (approved):
Great! 
Thanks

PR #4385, opened 2024-09-30
Fix config issues.

corentin-soriano:


nilsteampassnet (approved):

nilsteampassnet:
@corentin-soriano 
I just noticed that moving hidden field "encryptClientServer" place had an impact while user is not logged in.

Indeed when a user is not logged in then the this hidden field doesn't exist witch generated a warning popup.
I will revert the change from the index.php file.

nilsteampassnet:
Fixed with #4387

PR #4387, opened 2024-09-30
Fix for encryptClientServer input field

nilsteampassnet:
Fix for encryptClientServer input field


includes/js/functions.js
```diff
@@ -182,7 +182,7 @@ function prepareExchangedData(data, type, key, fileName = '', functionName = '')
             }
         }
     } else if (type === 'encode') {
-        if (parseInt($('#encryptClientServer').val()) === 0) {
+        if (parseInt($('#encryptClientServerStatus').val()) === 0) {
```
corentin-soriano, line 185: Well seen, I wanted to do that without success but I forgot to modify this line!

PR #4388, opened 2024-09-30
Add email field to admin account setup #4382

evertton:
I verified that the admin user creation query already had the "admin_email" field defined. [Here](https://github.com/nilsteampassnet/TeamPass/blob/bdd3fd9af5f6016ec6fbd07294d59ac3ecb560d4/install/install.queries.php#L842)

So I just did the html and js.

### DIFF
--- a/install/install.js
+++ b/install/install.js
@@ -97,6 +97,8 @@ function checkPage()
         }
     } else if (step === "4") {
     // STEP 4
+
+        // Password checks
         if ($("#admin_pwd").val() === "") {
             alertify
                 .error('<i class="fas fa-ban mr-2"></i>You must define a password for Administrator account.', 10)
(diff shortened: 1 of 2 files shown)


install/install.js
```diff
@@ -112,13 +114,26 @@ function checkPage()
...
-            multiple = "";
         }
+        
+        // Email checks
+        if ($("#admin_email").val() === "") {
+            alertify
+                .error('<i class="fas fa-ban mr-2"></i>You must define an email for Administrator account.', 10)
+                .dismissOthers();
+            return false;
+        } else if (!/^[^\s@]+@[^\s@]+\.[^\s@]+$/.test($("#admin_email").val())) {
```
corentin-soriano, line 125: This is not directly related to this change but it highlights it: The installation script seems to be vulnerable to SQL injection. The risk is mitigated since it is limited to the installation and we delete the file at the end.

Edit the request on the fly:
image
image

No php error but the table is empty:
Capture d'écran 2024-10-01 092453

I think other fields like saltkey might be vulnerables but need tests to validate.

Can you add email filter on the php side too?

evertton: I had also made these observations, but I had the idea of ​​doing the mapping first and suggesting the necessary corrections in another PR.

@nilsteampassnet, What do you think?

nilsteampassnet: @evertton 
For my point of view the PR should be complete.
But on the other hand, the vulnerability was already there.
So I will merge the PR and proovide a fix in order to fix the vulnerability in global

PR #4406, opened 2024-10-09
Fix role visibility display issue.

corentin-soriano:
My user role:
image

Add visibility on role:
image

Remove visibility and folder disappear successfully:
image

However I still see `'test` role with another user that has access to this folder:
image

This issue occur when 'type' field is empty:
```sql
mysql> SELECT * FROM teampass_roles_values WHERE type = "";
+--------------+---------+-----------+------+
| increment_id | role_id | folder_id | type |
+--------------+---------+-----------+------+
|            3 |       1 |         3 |      |
|            5 |       1 |         5 |      |
|          443 |       1 |       455 |      |
|          456 |       8 |        95 |      |
+--------------+---------+-----------+------+
4 rows in set (0.00 sec)
```

### DIFF
--- a/install/upgrade_run_3.1.php
+++ b/install/upgrade_run_3.1.php
@@ -588,8 +588,14 @@ if (file_exists($configFilePath)) {
     }    
 }
 
-
 //---<END 3.1.2
+
+
+//--->BEGIN 3.1.3
+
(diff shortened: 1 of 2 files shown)


nilsteampassnet (approved):
Nice

Issue #4404, opened 2024-10-08
Fixing error in SQL syntax on /api/index.php/item/get?id=xxx


ego-kyriarchos:
Hi, I fixed the item Module on Teampass REST API.

### The query:
curl -s -X GET -k 'https://teampass.lan/api/index.php/item/get?id=841' -H "Authorization: Bearer ${TOKEN}"

### Error on php:
```bash
tail -f /var/log/php-fpm/www-error.log
[08-Oct-2024 15:00:56 UTC] PHP Fatal error:  Uncaught Exception: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')) ORDER BY i.id ASC' at line 3 in /var/www/html/teampass/api/Model/Database.php:51
Stack trace:
#0 /var/www/html/teampass/api/Model/ItemModel.php(47): Database->select()
#1 /var/www/html/teampass/api/Controller/Api/ItemController.php(221): ItemModel->getItems()
#2 /var/www/html/teampass/api/inc/bootstrap.php(59): ItemController->getAction()
#3 /var/www/html/teampass/api/index.php(91): itemAction()
#4 {main}
  thrown in /var/www/html/teampass/api/Model/Database.php on line 51
```

### Solution on
/var/www/html/teampass/api/Controller/Api/ItemController.php

```php
	$sql_constraint = ' AND (i.id_tree IN ('.$userData['folders_list'].')'; #OR i.id IN ('.$userData['restricted_items_list'].'))';
	if (!empty($userData['restricted_items_list'])) {
		$sql_constraint = $sql_constraint.'OR i.id IN ('.$userData['restricted_items_list'].')';
	}
	$sql_constraint = $sql_constraint.')';
```

nilsteampassnet:
Thank you

PR #4408, opened 2024-10-09
Fixing error in SQL syntax on /api/index.php/item/get?id=xxx

nilsteampassnet:
Merged suggestion from #4404

### DIFF
--- a/includes/config/include.php
+++ b/includes/config/include.php
@@ -28,7 +28,7 @@
 
 define('TP_VERSION', '3.1.2');
 define("UPGRADE_MIN_DATE", "1727110744");
-define('TP_VERSION_MINOR', '5');
+define('TP_VERSION_MINOR', '8');
 define('TP_TOOL_NAME', 'Teampass');
 define('TP_ONE_DAY_SECONDS', 86400);
 define('TP_ONE_WEEK_SECONDS', 604800);
--- a/includes/config/include.php
+++ b/includes/config/include.php
@@ -28,7 +28,7 @@
 
 define('TP_VERSION', '3.1.2');
 define("UPGRADE_MIN_DATE", "1727110744");
-define('TP_VERSION_MINOR', '134');
+define('TP_VERSION_MINOR', '5');
 define('TP_TOOL_NAME', 'Teampass');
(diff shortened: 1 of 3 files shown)


includes/config/include.php
```diff
@@ -28,7 +28,7 @@
 
 define('TP_VERSION', '3.1.2');
 define("UPGRADE_MIN_DATE", "1727110744");
-define('TP_VERSION_MINOR', '134');
+define('TP_VERSION_MINOR', '5');
```
corentin-soriano, line 31: 3.1.2.5 ?

corentin-soriano:
I think there is another issue:
image

nilsteampassnet:
> I think there is another issue

Not sure to understand.

corentin-soriano:
> > I think there is another issue
> 
> Not sure to understand.

I don't see any call to getAction(). I think @ego-kyriarchos, added the missing call too.

nilsteampassnet:
> > > I think there is another issue
> > 
> > 
> > Not sure to understand.
> 
> I don't see any call to getAction(). I think @ego-kyriarchos, added the missing call too.

It is used through `bootstrap.php` file.
You have instruction: `$strMethodName = $actions[0] . 'Action';`
with `$actions[0]` coming from the URL, and in this case the value would be `get`.
So  `$strMethodName = 'getAction';` which is the name of the called function

corentin-soriano:
> > > > I think there is another issue
> > > 
> > > 
> > > Not sure to understand.
> > 
> > 
> > I don't see any call to getAction(). I think @ego-kyriarchos, added the missing call too.
> 
> It is used through `bootstrap.php` file. You have instruction: `$strMethodName = $actions[0] . 'Action';` with `$actions[0]` coming from the URL, and in this case the value would be `get`. So `$strMethodName = 'getAction';` which is the name of the called function

Thank you for the clarification!

corentin-soriano (approved):
Can you push the TP_MINOR_VERSION before merging?

nilsteampassnet:
The minor version will be fixed directly under master branch as it is a counter based upon the number of commits inside the branch ;)

PR #4414, opened 2024-10-11
Update spanish.php

Pruebas77:
Translations from the English file are added, and the translations are homogenized.


nilsteampassnet:
I have updated [POEditor ](https://poeditor.com/projects/po_edit?id_language=152&per_page=20&id=433631)

PR #4431, opened 2024-10-23
Fix user password expiration days error

FengZh61:
Missing judgment condition in code handling pw_life_duration>0

Fix type error in pw_life_duration and user-last_pw_change values

```
PHP Fatal error:  Uncaught TypeError: Unsupported operand types: int - string in /var/www/html/teampass/sources/core.php:542\nStack trace:\n#0 /var/www/html/teampass/index.php(110): require_once()\n#1 {main}\n  thrown in /var/www/html/teampass/sources/core.php on line 542
```

### DIFF
--- a/sources/core.php
+++ b/sources/core.php
@@ -539,7 +539,7 @@ if (isset($SETTINGS['ldap_mode']) === true && (int) $SETTINGS['ldap_mode'] === 1
             $session->set('user-num_days_before_exp', 'infinite');
             $session->set('user-validite_pw', 1);
         } else {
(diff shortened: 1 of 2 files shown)


FengZh61:
@corentin-soriano @nilsteampassnet 

When a user's password expires, the expected effect is to jump directly to the page for changing the password when logging in, similar to the function of changing the password after the user has successfully logged in, by entering the current password, a new password, and repeating the new password.
The page is written in a way that I don't really understand, can you schedule a solution to this problem?

nilsteampassnet:
@FengZh61
Thank you
What a coincidence, this morning I have identified the bug regarding this status.
I'm currently working on a fix.
I will not merge as is your pr but will reuse it, you put me in an interesting direction.

FengZh61:
> @FengZh61 Thank you What a coincidence, this morning I have identified the bug regarding this status. I'm currently working on a fix. I will not merge as is your pr but will reuse it, you put me in an interesting direction.

Great, I'll close this pull request and wait for your fix.

Issue #4422, opened 2024-10-17
Errors editing after import


rcruz-pntlb:
Hi! all, 

I detected some malfunction when importing from CSV file.  Thanks in advance for your help/advice/hints ;-)... so here we go,

Describing general config:
* Ubuntu 20.04.6 LTS
* PHP version: 8.1.29
* Current Teampass version loaded 3.1.2.52
* Settings: ["Import from KeePass or CSV files enabled":yes,"Managers may edit or delete any items they can view":yes, "New sub-folders inherit permissions from parent folder":no, "Allow restricting items to users and roles": yes, "Can create a folder at root level": yes,"Folders may have duplicate names":yes]
* Existing Role "IT" assigned to Administrator users includes rights for adding/editing/deleting on any folder.
* Web client: Chrome versión 129.0.6668.101 (Build oficial) (64 bits) / Windows 11 23H2

### Steps to reproduce
1. From a user with IT role, I create a new directory, specifying ["Test Label","Root", "Weak","",""] and save.
2. From import option I choose a testing file, content as follows:
label,login,password,url,comments
"Logan","MyLogan","YoupiCheap","http://www.logan.fr",""
"Logan2","MyLogan2","YoupiCheap2","http://www.logan2.fr",""
"Logan3","MyLogan3","YoupiCheap3","http://www.logan3.fr",""
"Logan4","MyLogan4","YoupiCheap4","http://www.logan4.fr",""
"Logan5","MyLogan5","YoupiCheap5","http://www.logan5.fr",""
"Logan6","MyLogan6","YoupiCheap6","http://www.logan6.fr",""
4. And additional parameters ["Any user could edit all imported items":checked, "Any user part of your Groups could edit all imported items":checked], target folder = "Test Label", All Items Selected ... and clicking on perform ...
5. After import csv file, the entries are apparently in their place, with right data on it. Alright!... but when I try to edit any item and save it, a red tooltip appears on chrome showing following message: "No changes was performed". And still more, from this point, if I try to manually add new items to the same folder, the same error persists when I try to save it...  

[by making use of Chrome Developer Tools]: I think, but I'm not sure about, that the Update Action (when I edit some item) is not sent to server when I click on Save button. It seems that javascript flow breaks off, registering following error at Console level (IMPORTANT! not when "edit" button is clicked, but when previously entering to item for reading data):

[...]

PR #4435, opened 2024-10-29
Fix issue #4422: Resolve edit errors on previously CSV-imported items

rcruz-pntlb:
This pull request addresses a bug where items, after being imported via CSV, could not be edited correctly. The issue was due to missing role validation in the HTML rendering process, and it specifically affects the `pages/items.js.php` script.

Changes include:
- Added a role existence check before adding HTML badge elements.
- Refactored the HTML generation loop to ensure robustness if roles are missing in `data.roles_list`.
- Fixed a minor error in the code that was not detected in the previous commit.

Original issue: https://github.com/nilsteampassnet/TeamPass/issues/4422

### DIFF
--- a/pages/items.js.php
+++ b/pages/items.js.php
@@ -5463,8 +5463,7 @@ $var['hidden_asterisk'] = '<i class="fa-solid fa-asterisk mr-2"></i><i class="fa
                 // Show restrictions with Badges
                 var html_restrictions = '';
                 $.each(store.get('teampassItem').id_restricted_to, function(i, value) {
-                    html_restrictions +=https://github.com/rcruz-pntlb/TeamPass/edit/bugfix/4422-fix-item-edit-after-csv-import/pages/items.js.php
-                        '<span class="badge badge-info mr-2 mb-1"><i class="fa-solid fa-group fa-sm mr-1"></i>' +
+                    html_restrictions +='<span class="badge badge-info mr-2 mb-1"><i class="fa-solid fa-group fa-sm mr-1"></i>' +
                         data.users_list.find(x => x.id === parseInt(value)).name + '</span>';
                 });
                 /*BEGIN modification for issue #4422 bugfix/4422-fix-item-edit-after-csv-import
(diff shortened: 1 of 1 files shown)


pages/items.js.php
```diff
@@ -5463,15 +5463,28 @@ function(data) {
                 // Show restrictions with Badges
                 var html_restrictions = '';
                 $.each(store.get('teampassItem').id_restricted_to, function(i, value) {
-                    html_restrictions +=
+                    html_restrictions +=https://github.com/rcruz-pntlb/TeamPass/edit/bugfix/4422-fix-item-edit-after-csv-import/pages/items.js.php
                         '<span class="badge badge-info mr-2 mb-1"><i class="fa-solid fa-group fa-sm mr-1"></i>' +
```
corentin-soriano: I think this is a mistake.

pages/items.js.php
```diff
@@ -5463,15 +5463,28 @@ function(data) {
...
+                    @Author:     rcruz@pintaluba.com
+                    @Date:       10/29/2024
+
+                --REPLACES ORIGINAL CODE--
                 $.each(store.get('teampassItem').id_restricted_to_roles, function(i, value) {
                     html_restrictions +=
                         '<span class="badge badge-info mr-2 mb-1"><i class="fa-solid fa-group fa-sm mr-1"></i>' +
                         data.roles_list.find(x => x.id === parseInt(value)).title + '</span>';
+                });                                                
+                */
```
corentin-soriano, line 5480: Can you remove the old code instead of comment it please?

corentin-soriano (changes requested):

pages/items.js.php
corentin-soriano: Too fast, it's already fixed!

pages/items.js.php
rcruz-pntlb: Hi @corentin-soriano 
ok!! ... and the rest of comments, better to remove, too?

corentin-soriano: Comments should describe the associated code. Dates/issue links and others informations are in the git history!

PR #4447, opened 2024-11-06
Exclude personal folder from identAdmin function (admins can't have personal folders).

corentin-soriano:
High latencies in the back office if the administrator account has a personal folder declared in the nested_tree table.
Since administrators do not have access to items, this block of code seems useless.


nilsteampassnet (approved):
Nice 👍

PR #4449, opened 2024-11-07
Improve users_build_cache_tree scheduling.

corentin-soriano:
Avoid `user_build_cache_tree` duplicates on each folder creation.
Do not start the same task again if the scheduler is launched several times.

An interesting case where the task caused an outage: multiple user_build_cache_trees per user and 455 tasks were performing the same actions at the same time.
```bash
[root@localhost ~]# ps aux | grep background_tasks___user_task.php | wc -l
455
```

image

Yes, the server has reached 300 of load avg!


nilsteampassnet (approved):
Very nice.
I already experimented such outage but couldn't fixed it.
👍

PR #4453, opened 2024-11-09
Improved last_cron_exec parameter in misc table

nilsteampassnet:
Added a step for removing duplicated entries in misc table (during upgrade) 
Added a new index on type and intitule in misc table 
Simplified the last_cron_exec parameter update


includes/config/include.php
```diff
@@ -87,6 +87,7 @@
 define('DEBUGLDAP', false); //Can be used in order to debug LDAP authentication
 
 define("ADMIN_VISIBLE_OTP_ON_LDAP_IMPORT", true);
+define('SERVER_HAS_INTERNET_ACCESS', false);
```
corentin-soriano, line 90: This is a good idea, however it is not used anywhere currently.

install/upgrade_run_3.1.php
```diff
@@ -588,6 +588,52 @@
...
+    $deleteQuery = "DELETE FROM ".$pre."misc WHERE increment_id IN ($idsToDeleteStr)";
+    if (!mysqli_query(
+        $db_link,
+        $deleteQuery
+    )) {
+        echo '[{"finish":"1", "msg":"", "error":"An error appears when removing duplicated entries is_encrypted to table misc! ' . mysqli_error($db_link) . '!"}]';
+        mysqli_close($db_link);
+        exit();
+    }
+}
```
corentin-soriano, line 624: I think we could do the delete directly to avoid loops:
```php
$sql = "
DELETE FROM ".$pre."misc
WHERE increment_id IN (
    SELECT increment_id
    FROM ".$pre."misc AS t1
    WHERE EXISTS (
        SELECT 1
        FROM ".$pre."misc AS t2
        WHERE t1.type = t2.type
        AND t1.intitule = t2.intitule
        AND t1.increment_id < t2.increment_id
    )
)
";
```

sources/scheduler.php
```diff
@@ -47,22 +47,15 @@
 $SETTINGS = $configManager->getAllSettings();
 
 // Update row if exists
-$updated = DB::update(
-    prefixTable('misc'),
-    ['valeur' => time()],
-    'type = "admin" AND intitule = "last_cron_exec"'
+DB::insertUpdate(
```
corentin-soriano, line 50: I didn't know about `insertUpdate()`, it's more readable that way!

PR #4460, opened 2024-11-14
Use random encryption key instead and reset session id on login.

corentin-soriano:
Using the PHPSESSID cookie value as encryption key removes the benefit of the "httpOnly" cookie flag.
Good practice is to change the session id after successful authentication.


nilsteampassnet (commented):
Very nice 👍

PR #4465, opened 2024-11-14
Crypt session

nilsteampassnet:
Improves the session manager by encrypting all session files content

IMPORTANT: once files are copied on the server, it is required to remove all session files.
This deletion has to be performed before browsing Teampass again.

As an example: 
* Folder could be `/var/lib/php/sessions`
* Command to remove files: `sudo find /var/lib/php/sessions -type f -exec rm -f {} +`


corentin-soriano (approved):
It works fine, thank you!

PR #4474, opened 2024-11-19
Clean XSS before html encode on csv and keepass imports.

corentin-soriano:
Before this change:
image
After:
image

Note that visible payloads cannot be executed.


nilsteampassnet (approved):
GReat and simple

PR #4483, opened 2024-11-22
fix: fix user email incorrect format

stanislavefremov:
sendMailToUser function gets email body in $post_body variable and unconditionally applies htmlspecialchars to it, which escapes html tags. It maybe unneeded as later sendMailToUser passes email body to $emailService->sendMail method which performs XSS check through sanitizeEmailBody method and applies htmlspecialchars if needed.


stanislavefremov:
htmlspecialchars in sendMailToUser function seems to be unnecessary

corentin-soriano (commented):
It seems to me that this function is used by `mail_me` (direct user input). I think wee can add `$antiXSS->clean()` to clear eventual XSS payload without break the html.

PR #4486, opened 2024-11-24
Fix backup restore process

nilsteampassnet:
Some servers cannot be customized in depth.
This PR restores standard upload parameters.
It fixes also the backup restore process providing more feedback (%).
Note: embends also a fix on log_datatables for ORDER BY instruction.

### DIFF
--- a/sources/main.functions.php
+++ b/sources/main.functions.php
@@ -1865,7 +1865,6 @@ function prepareFileWithDefuse(
     string $type,
     string $source_file,
     string $target_file,
-    array $SETTINGS,
     string $password = null
 ) {
     // Load AntiXSS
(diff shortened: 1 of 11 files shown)


sources/main.functions.php
```diff
@@ -1932,9 +1929,9 @@ function defuseFileEncrypt(
     } catch (CryptoException\WrongKeyOrModifiedCiphertextException $ex) {
         $err = 'wrong_key';
     } catch (CryptoException\EnvironmentIsBrokenException $ex) {
-        $err = $ex;
+        $err = print_r($ex, true);
     } catch (CryptoException\IOException $ex) {
-        $err = $ex;
+        $err = print_r($ex, true);
```
corentin-soriano, line 1934: Does this can leak server private informations to frontend (server saltkey, saltkey path or other)?

nilsteampassnet: I have improved this.

Thanks

PR #4488, opened 2024-11-25
Enforce password log on copy or shown action.

corentin-soriano:
- Remove item password from DOM to prevent unrecorded actions on passwords.
- Add generic function `getItemPassword()` to get item password client side.
- Use `getItemPassword()` every time the password is needed (copy, show, edit item).


nilsteampassnet:
Works nice

Issue #4479, opened 2024-11-21
Fresh Install failing.


hitenmandalia:
### Steps to reproduce

1.  Attempt to install fresh 
2.
3.

### Expected behaviour
Should be able to pass the check screen

### Actual behaviour
Nothing happens on screen.  Errors shown

### Server configuration
**Operating system**: Alpine Linux

**Web server:** Nginx

**Database:** MySQL 8

**PHP version:** 8.1

**Teampass version:** 

**Teampass configuration file:**

**Updated from an older Teampass or fresh install:**
PLEASE attach to this issue the file `/includes/config/tp.config.php`.

Install never goes past the Checks screen:

![image](https://github.com/user-attachments/assets/798e3afb-0a41-4688-9123-6f7e6d966ecd)
![image](https://github.com/user-attachments/assets/26370f7d-bd15-4daf-b9d9-afbec0b4bace)

``
2024-11-21 15:53:18 [21-Nov-2024 15:53:18] WARNING: [pool www] child 61 said into stderr: "NOTICE: PHP message: PHP Notice:  file_get_contents(): Read of 12288 bytes failed with errno=21 Is a directory in /var/www/html/vendor/teampassclasses/sessionmanager/src/SessionManager.php on line 44"
2024-11-21 15:53:18 [21-Nov-2024 15:53:18] WARNING: [pool www] child 61 said into stderr: "NOTICE: PHP message: PHP Stack trace:"
2024-11-21 15:53:18 [21-Nov-2024 15:53:18] WARNING: [pool www] child 61 said into stderr: "NOTICE: PHP message: PHP   1. {main}() /var/www/html/install/install.queries.php:0"
2024-11-21 15:53:18 [21-Nov-2024 15:53:18] WARNING: [pool www] child 61 said into stderr: "NOTICE: PHP message: PHP   2. require_once() /var/www/html/install/install.queries.php:57"
2024-11-21 15:53:18 [21-Nov-2024 15:53:18] WARNING: [pool www] child 61 said into stderr: "NOTICE: PHP message: PHP   3. TeampassClasses\SessionManager\SessionManager::getSession() /var/www/html/sources/main.functions.php:62"
2024-11-21 15:53:18 [21-Nov-2024 15:53:18] WARNING: [pool www] child 61 said into stderr: "NOTICE: PHP message: PHP   4. file_get_contents($filename = '/') /var/www/html/vendor/teampassclasses/sessionmanager/src/SessionManager.php:44"
[...]

PR #4491, opened 2024-11-25
Fix install

nilsteampassnet:
In relation to #4479


install/install.queries.php
```diff
@@ -290,17 +354,14 @@ function encryptFollowingDefuse($message, $ascii_key)
...
                 $tmp = mysqli_num_rows(mysqli_query($dbTmp, "SELECT * FROM `_install` WHERE `key` = 'url_path'"));
                 if (intval($tmp) === 0) {
-                    mysqli_query($dbTmp, "INSERT INTO `_install` (`key`, `value`) VALUES ('url_path', '" . $escapedUrlPath . "');");
+                    mysqli_query($dbTmp, "INSERT INTO `_install` (`key`, `value`) VALUES ('url_path', '" . $session_url_path . "');");
                 }
 
                 $tmp = mysqli_num_rows(mysqli_query($dbTmp, "SELECT * FROM `_install` WHERE `key` = 'absolute_path'"));
                 if (intval($tmp) === 0) {
-                    mysqli_query($dbTmp, "INSERT INTO `_install` (`key`, `value`) VALUES ('absolute_path', '" . $escapedAbsPath . "');");
+                    mysqli_query($dbTmp, "INSERT INTO `_install` (`key`, `value`) VALUES ('absolute_path', '" . $session_abspath . "');");
```
corentin-soriano, line 364: Can you add `mysqli_real_escape_string()` on `$session_url_path` and `$session_abspath` please?

install/install.queries.php
```diff
@@ -1497,8 +1508,8 @@ function encryptFollowingDefuse($message, $ascii_key)
 define("DB_CONNECT_OPTIONS", array(
     MYSQLI_OPT_CONNECT_TIMEOUT => 10
 ));
-define("SECUREPATH", "' . $securePath . '");
-define("SECUREFILE", "' . $filesecure. '");
+define("SECUREPATH", "' . $session_sk_path . '");
```
corentin-soriano, line 1511: Wouldn't a user who mistakenly inserted a quote into the form find themselves stuck with a HTTP 500 error?

install/install.queries.php
```diff
@@ -1464,7 +1470,12 @@ function encryptFollowingDefuse($message, $ascii_key)
                             unlink($filename);
                         }
                     }
-                    //echo ">". $inputData['db']['db_pw']." -- ".$new_salt." ;; ";
+
+                    //
```
corentin-soriano, line 1474: Unfinised comment!

corentin-soriano (changes requested):

install/install.queries.php
nilsteampassnet: Yep.
The whole install should be reworked, it is the same since years.
I will on this topic in next weeks.

install/install.queries.php
nilsteampassnet: You are right.
I will just add an str_replace but a full control should be done another way.

install/install.queries.php
nilsteampassnet: Yep

nilsteampassnet:
The fix provides minor changes to permit the installation.
Nevertheless a full rework of it should be performed in the future. This part is indeed quite old.

corentin-soriano (approved):
I just tested, I made some changes (I asked to merge into this branch): #4503
I had a few hangs due to mishandled cases in anti bruteforce, an error after removing the temporary securefile and minor modifications.

install/install.js
```diff
@@ -47,7 +39,9 @@ var global_error_on_query = false,
     dbInfo = "",
     index = "",
     multiple = "",
-    jsonValues = "";
+    jsonValues = "",
+    skFile = "";
```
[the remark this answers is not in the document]
nilsteampassnet: no, just that it was the case since the begining and did not readjust.
As said all this should be rewritten.

PR #4412, opened 2024-10-11
Code review fixes

nilsteampassnet:
Set of small fixes during code review

PR #4498, opened 2024-11-27
Fix items tasks error

corentin-soriano:
The scheduler is stuck on tasks that handle sharekeys creation:
- Fix SQL query:
```
#0 /var/www/teampass/vendor/sergeytsalkov/meekrodb/db.class.php(890): MeekroDB->queryHelper(Array, Array)
#1 /var/www/teampass/vendor/sergeytsalkov/meekrodb/db.class.php(116): MeekroDB->query('SELECT id, publ...', Array)
#2 /var/www/teampass/sources/main.functions.php(2640): DB::__callStatic('query', Array)
```
- Improve task recursion to avoid high CPU load.
- Error on item creation with empty password since #4412 (`if ($encrypted === false) {` -> `if (empty($encrypted)) {`).
- Error on tasks page (missing field arguments in SQL query).


nilsteampassnet (approved):
Reaaly nice 👍

PR #4505, opened 2024-11-29
Updated language files with new strings.

nilsteampassnet:


corentin-soriano (commented):
`bruteforce_wait` is the only missing!

corentin-soriano (approved):

PR #4512, opened 2024-12-04
Renewed installation process

nilsteampassnet:
Installation steps should be easier to maintain and change from now.
Reviewed forms for better user feedbacks.
Get usage of new versions of bootstrap and fontawesome, but not yet deployed on teampass.


corentin-soriano:
Great work, I didn't expect to see the PR finished so soon!

A few things to change before the review:
- Can you add bootstrap/fontawesome in another branch? My browser is crashing when displaying the PR as is!
image
- I don't find the install/install folder clear, I would have named the subfolder `tasks` or `steps`.
- I think you need to rebase this branch on `develop` and change the base branch in the PR because it's a feature.
- For versioning, should we change to 3.1.4.BETA1 or something like that so as not to have two branches on the "same" version but with different code?

Issue #4528, opened 2024-12-16
Passwords are not displayed in modern browsers due to synchronous AJAX requests


jlecoustre:
Description of the Problem

When accessing passwords in TeamPass using modern browsers (e.g., Chrome, Firefox, Android), the passwords are not displayed and cannot be copied. Instead, the browser's console shows multiple warnings related to synchronous AJAX requests being blocked. The application works correctly in older browsers.

The error messages in the browser console are as follows:

[Violation] Permissions policy violation: Synchronous requests are disabled by permissions policy.
jquery.min.js:2 [Violation] Permissions policy violation: Synchronous requests are disabled by permissions policy.
send @ jquery.min.js:2
ajax @ jquery.min.js:2
getItemPassword @ index.php?page=items:3735

Steps to Reproduce

    Open TeamPass in a modern browser (e.g., Chrome 117+, Firefox 117+, or an Android browser).
    Navigate to the items page and attempt to display or copy a password.
    Observe that no password is shown and the console displays the above warnings.

Expected Behavior

Passwords should be displayed and copied correctly in all browsers, regardless of platform or version.

Based on the error messages and investigation:

    The issue arises from the use of synchronous AJAX requests (e.g., async: false) in jquery.min.js or index.php?page=items. Synchronous requests have been deprecated in modern browsers due to their negative impact on performance and user experience.
    Permissions policies in modern browsers explicitly block synchronous requests in most scenarios.

Proposed Solution

To resolve the issue and ensure compatibility with modern browsers:

    Replace synchronous AJAX requests with asynchronous ones (async: true).
    Use Promises or async/await to handle asynchronous operations if needed.
    Update the getItemPassword function in index.php?page=items to remove dependencies on synchronous behavior.
    If applicable, update the clipboard.min.js file to align with modern best practices.

Temporary Workaround

While awaiting a fix, users can:

    Use older browsers that do not block synchronous requests (not recommended for security reasons).
    Run browsers with flags to disable security restrictions, such as --disable-web-security in Chrome (only for local testing, not in production).

Environment Details

    TeamPass Version: 3.1.3.1
    PHP Version: PHP 8.3.14 (fpm-fcgi)
    Browser: Chrome 117+, Firefox 117+, Android browsers.
    Operating Systems: Linux, Android, Windows

Additional Information

This issue is critical as it prevents users from accessing their passwords on modern browsers. Resolving it will improve compatibility and user experience across all platforms.

[...]

PR #4538, opened 2024-12-22
Fixing #4528

nilsteampassnet:
Main idea is to remove library ClipboardJS in order to permit usage of async calls when copying data to clipboard. The change relies on the usage of navigator.clipboard instruction.


corentin-soriano:
I'm going to try to look this week, however I think it will be better to merge it into develop because it seems more like an improvement than a fix and we risk having regressions given the number of modified lines.

nilsteampassnet:
The point of view is correct @corentin-soriano 
I will try to change this.

PR #4539, opened 2024-12-22
Feature/clipboard

nilsteampassnet:
Fixing #4528
Main idea is to remove library ClipboardJS in order to permit usage of async calls when copying data to clipboard. The change relies on the usage of navigator.clipboard instruction.


pages/items.js.php
```diff
@@ -384,33 +384,33 @@ function(teampassApplication) {
...
+                    // Autohide
+                    setTimeout(() => {
+                        $(this).removeClass('pwd-shown');
+                        $('#card-item-pwd')
+                            .html('<?php echo $var['hidden_asterisk']; ?>')
+                            .removeClass('pointer_none');
+                        $('.pwd-show-spinner')
+                            .removeClass('fa-solid fa-eye fa-beat-fade text-warning')
+                            .addClass('fa-regular fa-eye');
+                    }, <?php echo isset($SETTINGS['password_overview_delay']) === true ? $SETTINGS['password_overview_delay'] * 1000 : 4000; ?>);
```
corentin-soriano, line 411: Is `$SETTINGS['password_overview_delay']` an int?

corentin-soriano (commented):
I haven't tested it yet, I'll take care of it as soon as I can with the changes to the install!

corentin-soriano:
I encounter an error if I copy an empty password from the list of items:
image
If I click on the button to display the password on an item without password, it is then broken for all other items
 (click on asterisks show it but not the eye):
image
The clipboard is cleaned instantly while the auto-clear feature is disabled:
image

I've noticed cases where the "copied to clipboard" message is not displayed on the copy of a username or URL. This usually happens when I open one item, then another.

nilsteampassnet:
@corentin-soriano thank you
The fix should handle this case.
Tested with empty password copied from items list
image

I could not reproduce the other point.
If no password then we should not have the "copy" button
image

I will have an eye on this point.

pages/items.js.php
```diff
@@ -408,7 +408,7 @@ function(teampassApplication) {
                         $('.pwd-show-spinner')
                             .removeClass('fa-solid fa-eye fa-beat-fade text-warning')
                             .addClass('fa-regular fa-eye');
-                    }, <?php echo isset($SETTINGS['password_overview_delay']) === true ? $SETTINGS['password_overview_delay'] * 1000 : 4000; ?>);
+                    }, <?php echo isset($SETTINGS['password_overview_delay']) && is_int($SETTINGS['password_overview_delay']) ? $SETTINGS['password_overview_delay'] * 1000 : 4000; ?>);
```
corentin-soriano, line 411: I think we need to check if the setting is greater than 0 too.
image

corentin-soriano (changes requested):
To reproduce the other issue, I just open an item with empty password and click on the eye:
image
Then the button doesn't show the password on all other items without refresh teampass:
image

pages/items.js.php
nilsteampassnet: Hi @corentin-soriano 
Didn't catched the show icon from the start but ok I could reproduce.
I've changed the logic. If no password, an igllo icon is shown informing the user that no pwd exists for this item.

pages/search.js.php
```diff
@@ -289,90 +289,135 @@ function(data) {
...
-                                    progressBar: true
-                                }
-                            );
+
+                        // Copy to clipboard
+                        await navigator.clipboard.writeText(password);
+
+                        // Notification for the user
+                        const clipboardDuration = store.get('teampassSettings').clipboard_life_duration || 0;
+                        if (clipboardDuration === 0) {
```
corentin-soriano, line 312: `clipboardDuration` is a string so `clipboardDuration === 0` is always false.

pages/items.js.php
```diff
@@ -4502,6 +4435,105 @@ function(teampassApplication) {
...
+                            // A password can be empty. Just exit.
+                            return;
+                        }
+
+                        // Copy the password to the clipboard
+                        await navigator.clipboard.writeText(password);
+
+                        // User notifications
+                        const clipboardDuration = store.get('teampassSettings').clipboard_life_duration || 0;
+                        if (clipboardDuration === 0) {
```
corentin-soriano, line 4461: `clipboardDuration` is a string so `clipboardDuration === 0` is always false.

pages/items.js.php
```diff
@@ -5191,55 +5246,71 @@ function(teampassItem) {
...
-                                            progressBar: true
-                                        }
-                                    );
+
+                                // Copy to clipboard
+                                await navigator.clipboard.writeText(password);
+
+                                // Notification for the user
+                                const clipboardDuration = store.get('teampassSettings').clipboard_life_duration || 0;
+                                if (clipboardDuration === 0) {
```
corentin-soriano, line 5267: `clipboardDuration` is a string so `clipboardDuration === 0` is always false.

corentin-soriano (changes requested):
I noticed one last issue on search page:
image

pages/search.js.php
```diff
@@ -289,90 +289,135 @@ function(data) {
...
+                        const password = await getItemPassword('at_password_copied', 'item_id', $('#btn-copy-pwd').data('id'));
 
-                            return item_pwd;
+                        if (!password) {
+                            toastr.error('<?php echo $lang->get("error_fetching_password"); ?>', '', {
+                                timeOut: 3000,
+                                positionClass: 'toast-bottom-right',
+                                progressBar: true
+                            });
+                            return;
```
corentin-soriano, line 304: Here we have a double error message:
image

I think the `error_fetching_password` here can be removed and we can directly do the `return`.

pages/search.js.php
```diff
@@ -289,90 +289,135 @@ function(data) {
...
+                        toastr.error('<?php echo $lang->get("clipboard_error"); ?>', '', {
+                            timeOut: 3000,
+                            positionClass: 'toast-bottom-right',
+                            progressBar: true
+                        });
+                    }
+                });
+                
+                // Click handler to copy the login
+                document.getElementById('btn-copy-login').addEventListener('click', async function() {
```
corentin-soriano, line 352: I noticed that `#btn-copy-login` is missing on items without a login.
This causes an error and the search spinner is never removed:
image

corentin-soriano (changes requested):

corentin-soriano (approved):
Looks good to me!

PR #4546, opened 2024-12-26
Fix restricted_to issue on item creation.

corentin-soriano:
Will fix #4542


DobroFenix:
Thank you. It really works and helped me a lot.

nilsteampassnet (approved):

PR #4551, opened 2024-12-31
Code clean

nilsteampassnet:
- Updating php libraries with composer
- Several fixes based upon Scrutinizer analysis


corentin-soriano (commented):
I agree with these changes but I would prefer them to be added in the develop branch because they are improvements and not corrections.

nilsteampassnet:
@corentin-soriano I see your point of view.
The original idea was to update the library tcpdf as a vulnerability has been identified in the used version.
But anyhow, the changes performed on could indeed being seen as "improvements".
I will switch branch

PR #4584, opened 2025-02-08
Fix issue while creating new user through oauth2

nilsteampassnet:
oAuth2 new user creation could fails.
This fix permits to ensure the creation is done properly.


corentin-soriano (approved):
Looks good!

PR #4700, opened 2025-04-27
Feature/oauth2 user management

nilsteampassnet:
Provide same configuration for LDAP and oAuth2 users.
This evolution forces oAuth2 users to be created from Users page before first authentication in Teampass.

### DIFF
--- a/pages/users.js.php
+++ b/pages/users.js.php
@@ -86,7 +86,7 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
         constVisibleOTP = false,
         userClipboard,
         ProcessInProgress = false,
-        debugJavascript = true;
+        debugJavascript = false;
 
     browserSession(
         'init',
--- a/pages/users.js.php
+++ b/pages/users.js.php
@@ -86,7 +86,7 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
         constVisibleOTP = false,
         userClipboard,
         ProcessInProgress = false,
-        debugJavascript = false;
+        debugJavascript = true;
 
     browserSession(
         'init',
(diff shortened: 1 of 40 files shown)


pages/users.js.php
```diff
@@ -2148,14 +2248,135 @@ function(data) {
...
+                                '</td>' +
+                                '<td>' +
+                                '<i class="fa-solid fa-info-circle ml-3 infotip text-info pointer text-center" data-toggle="tooltip" data-html="true" title="' +
+                                '<p class=\'text-left\'><i class=\'fas fa-user mr-1\'></i> ' +
+                                user.displayName + '</p>' +
+                                '<p class=\'text-left\'><i class=\'fas fa-envelope mr-1\'></i>' + (user.mail !== null ? '' + user.mail + '' : '') + '</p>' +
+                                '"></i>' +
+                                '</td><td>' +
+                                (user.userInTeampass === 0 ? '' :
+                                '<i class="fa-solid ' + (user.userAuthType === 'oauth2' ? 'fa-toggle-on text-info ' : 'fa-toggle-off ') + 'mr-1 text-center pointer action-change-oauth2-synchronization" data-user-id="' + user.userInTeampass + '" data-user-auth-type="' + user.userAuthType + '" infotip title="<?php echo $lang->get('toggle_user_authentifcation'); ?>"></i>') +
```
corentin-soriano, line 2305: `toggle_user_authentication`

pages/users.js.php
```diff
@@ -2148,14 +2248,135 @@ function(data) {
...
+
+                    // Build list box of new roles that could be created
+                    $('#oauth2-new-role-selection')
+                        .empty()
+                        .append('<option value="">--- <?php echo $lang->get('select'); ?> ---</option>');
+                    $.each(data.ad_groups, function(i, group) {
+                        tmp = data.teampass_groups.filter(p => p.title === group);
+                        if (tmp.length === 0) {
+                            $('#oauth2-new-role-selection').append(
+                                '<option value="' + group + '">' + group + '</option>'
```
corentin-soriano, line 2354: Is `group` XSS safe?

pages/users.js.php
```diff
@@ -2148,14 +2248,135 @@ function(data) {
...
+                                '<td>' + user.login +
+                                '</td>' +
+                                '<td>' +
+                                '<i class="fa-solid fa-info-circle ml-3 infotip text-info pointer text-center" data-toggle="tooltip" data-html="true" title="' +
+                                '<p class=\'text-left\'><i class=\'fas fa-user mr-1\'></i> ' +
+                                user.displayName + '</p>' +
+                                '<p class=\'text-left\'><i class=\'fas fa-envelope mr-1\'></i>' + (user.mail !== null ? '' + user.mail + '' : '') + '</p>' +
+                                '"></i>' +
+                                '</td><td>' +
+                                (user.userInTeampass === 0 ? '' :
```
corentin-soriano, line 2304: This `html` block appears to bring multiple XSS vulnerabilities with HTML concatenations with variables.

corentin-soriano (commented):
I think the logic should be reversed: why add more work for the administrator?

It would be better to offer the administrator the option of limiting access to a specific role that would be provided in the OAuth claims.
When creating the account on the OAuth provider, the administrator sets the role to the account to allow access to TeamPass.

PR #4707, opened 2025-05-04
Fix for error shown on login page

### LINKED ISSUE #4706: Malformed UTF-8 data
I know this thread is mentioned here [https://github.com/nilsteampassnet/TeamPass/issues/4698](https://github.com/nilsteampassnet/TeamPass/issues/4698) and it has been closed, but I too got this errors **Malformed UTF-8 data**

- New install on Ubuntu 20.04
- I did follow the troubleshoot thread
- I did reinstall TeamPass several times using different methods and procedures (directory, chmod, http, https, db type. . .)

Nothing worked

It would be nice to read a solution

nilsteampassnet:
* Should fix #4706
* Prevent bad handling of special characters in user password

### DIFF
--- a/sources/identify.php
+++ b/sources/identify.php
@@ -210,7 +210,7 @@ function identifyUser(string $sentData, array $SETTINGS): bool
         $SETTINGS,
         (string) $server['PHP_AUTH_USER'],
         (string) $server['PHP_AUTH_PW'],
-        (string) filter_var($dataReceived['pw'], FILTER_SANITIZE_FULL_SPECIAL_CHARS),
+        (string) $dataReceived['pw'],   // THis is a password, so no need to escape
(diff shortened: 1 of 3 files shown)


sources/identify.php
```diff
@@ -210,7 +210,7 @@ function identifyUser(string $sentData, array $SETTINGS): bool
         $SETTINGS,
         (string) $server['PHP_AUTH_USER'],
         (string) $server['PHP_AUTH_PW'],
-        (string) filter_var($dataReceived['pw'], FILTER_SANITIZE_FULL_SPECIAL_CHARS),
+        (string) $dataReceived['pw'],   // THis is a password, so no need to escape
```
corentin-soriano, line 213: @nilsteampassnet This is not backward compatible and will break existing accounts.

PR #4712, opened 2025-05-06
Fix multiple roles permission conflicts

FengZh61:
When a user has multiple roles for a folder, the permissions are deduplicated, and when both read and write permissions are available, the write permissions will be retained.


FengZh61:
It seems that only read and write permissions are taken into account, and other permissions are not taken into account. How to define the priority of permissions?

Issue #4693, opened 2025-04-24
Failed to import XML


rafinhamvieira:
Hello @nilsteampassnet  !

I had upgrade the Teampass for version 3.1.4.15, and I can import my passwords, but when I click in "execute", It show Reading file and never pass this, follow the attachment.

![Image](https://github.com/user-attachments/assets/ba670769-39cf-4c6e-a104-920c1beff631)

Image

Steps to reproduce
Import XML or CSV file.

Server configuration
Operating system: Ubuntu 24.04.2 LTS

Web server: Apache/2.4.58

Database: mariadb Ver 15.1 Distrib 10.11.11-MariaDB, for debian-linux-gnu (x86_64)

PHP version: PHP 8.4.6

Teampass version: version 3.1.4.12.

Teampass configuration variables:

rafinhamvieira:
Hello, any return? I need import my passwords.

rafinhamvieira:
In error log show me this 

[...]

PR #4744, opened 2025-06-04
Fix for #4693 - Problem with import XML

rafinhamvieira:
I updated the code for fix the bug when i try to import XML from keepass

### DIFF
--- a/sources/import.queries.php
+++ b/sources/import.queries.php
@@ -6,19 +6,19 @@ declare(strict_types=1);
  * Teampass - a collaborative passwords manager.
  * ---
  * This file is part of the TeamPass project.
- * 
+ *
  * TeamPass is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
  * the Free Software Foundation, version 3 of the License.
- * 
+ *
  * TeamPass is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  * GNU General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see <https://www.gnu.org/licenses/>.
- * 
+ *
  * Certain components of this file may be under different licenses. For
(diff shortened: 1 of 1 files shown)


sources/import.queries.php
```diff
@@ -675,7 +675,7 @@ function detectFileEncoding($filepath): string
                         'pw' => $itemPassword,
                         'pw_iv' => '',
                         'url' => empty($item['url']) === true ? '' : substr($item['url'], 0, 500),
-                        'id_tree' => is_null($item['folder_id']) === true ? $targetFolderId : (int) $item['folder_id'],
+                        'id_tree' => isset($post_folders[$item['parentFolderId']]['id']) ? (int)$post_folders[$item['parentFolderId']]['id'] : 0,
```
nilsteampassnet, line 678: Variable $post_folders is not defined in this case
I don't the interest of this change as it will break the execution.

sources/import.queries.php
```diff
@@ -512,7 +512,7 @@ function detectFileEncoding($filepath): string
                     // Insert folder and get its ID
                     $params = [
                         'title' => (string) $currentFolder,
-                        'parent_id' => (int) $parentId,
+                        'parent_id' => isset($parentId) && $parentId !== null ? $parentId : 0,
```
nilsteampassnet, line 515: Test is correct.
Missing the (int)

sources/import.queries.php
```diff
@@ -1270,7 +1270,7 @@ function createFolder($folderTitle, $parentId, $folderLevel, $startPathLevel, $l
         DB::insert(
             prefixTable('nested_tree'),
             array(
-                'parent_id' => $parentId,
+                'parent_id' => isset($parentId) && $parentId !== null ? $parentId : 0,
```
nilsteampassnet, line 1273: Test is correct.
Missing the (int)

nilsteampassnet (changes requested):
The commit is not signed

sources/import.queries.php
nilsteampassnet, line 1273: Not needed as $parentId is set by function parameter.
As a consequence, it is always defined.

PR #4752, opened 2025-06-06
Displaying error toast if fetch user fails

hmeza:


pages/users.js.php
```diff
@@ -133,13 +134,17 @@
         'stateSave': true,
         'autoWidth': true,
         'ajax': {
-            url: '<?php echo $SETTINGS['cpassman_url']; ?>/sources/users.datatable.php',
+            url: '<?php $SETTINGS['cpassman_url']= ''; echo $SETTINGS['cpassman_url']; ?>/sources/users.datatable.php',
```
corentin-soriano, line 137: This assignment looks like a bug.
This value comes from the back office configuration and should not be reassigned.

corentin-soriano (changes requested):

PR #4753, opened 2025-06-06
Displaying error toast if fetch user fails

hmeza:


hmeza:
I added a variable to control the Loading toast as I had 2 toasts when accessing this screen, so I wanted to control which toast was to be removed.

nilsteampassnet:
Thank you @hmeza
It is merged

PR #4799, opened 2025-07-09
enhancement: add MD5 hash validation to file integrity check

gudmmk:
## What does this PR do?
- Enhances the existing file integrity checker to validate each file’s content by comparing MD5 hashes against a reference list.
- Ensures that modifications to file contents are detected, even if filenames remain unchanged.

## Why?
- Previously, the integrity check only compared filenames to files_reference.txt, which could not catch content tampering.
- Now it verifies actual file contents, improving security and trust in the deployment.

## How was this tested?
- Intentionally modified files in the deployment directory.
- Verified that the admin interface flagged hash mismatches, displaying expected vs. actual hashes.

## Notes
- The format of files_reference.txt now expects:
 `relative/path/to/file.md5hash`
For example:
`vendor/autoload.php 1bc29b36f623ba82aaf6724fd3b16718`

- A helper script is provided at scripts/task_generate_files_reference.php to generate this file in the correct format.
Run it from the TeamPass project root:
`php scripts/task_generate_files_reference.php`
- This means to adopt the new integrity check, you just regenerate files_reference.txt using this script.

## Security Considerations
- Currently, files_reference.txt is stored in the same webroot. This means if the server is compromised, an attacker could update this file to hide malicious changes.
- A more robust approach would be to fetch the reference from the official TeamPass repository (or store it outside the webroot). This would add complexity (different TeamPass versions have different files), so this PR keeps to the existing local approach.

### DIFF
--- a/scripts/task_generate_files_reference.php
+++ b/scripts/task_generate_files_reference.php
@@ -1,0 +1,89 @@
+#!/usr/bin/php
+<?php
+
+/**
+ * task_generate_files_reference.php
+ * ---
+ * Helper script to generate a `files_reference.txt` file
+ * listing all project files and their MD5 hashes.
+ * 
+ * Output format:
+ *     <relative/path/to/file> <md5_hash>
+ * 
+ * Example:
+ *     vendor/autoload.php 1bc29b36f623ba82aaf6724fd3b16718
+ *
+ * This is used by the enhanced TeamPass file integrity check.
+ *
+ * Usage:
+ *     php scripts/task_generate_files_reference.php
+ *
+ * Configuration:
+ * - Adjust `$excludeDirs` to skip entire directories.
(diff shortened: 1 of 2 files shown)


nilsteampassnet:
@gudmmk 
Thank you for your input.
Currently the reference file is not generated using a script.
I'm using the ls-file command from git.
It permits to get the exact list of files inside the github project which is the goal here.
Getting the md5 of the files is a really good suggestion, but it can be generated out without a script
I will change the existing command in the release script to : `git ls-files --cached -z | xargs -0 md5sum > files_reference.txt`
By this it will include the md5 signature.
Just need to ensure the evolution you provided fits the output format ;)

nilsteampassnet:
It works smoothly, thank you @gudmmk 
I had adapted the command to fit the format.
I used your PR to also change the way the results are shown.

PR #4801, opened 2025-07-10
Fix: upload any extensions file not being honored

gudmmk:
### Summary
- Fixes an issue where the “Upload any extension file” setting (upload_all_extensions_file) was not being respected by the file upload scripts. Even when enabled in the TeamPass UI, uploads were still blocked if the file extension wasn’t explicitly listed in the allowed types.

### Changes
- Updated upload.attachments.php and upload.files.php to check the upload_all_extensions_file setting before enforcing extension validation.
- When this setting is enabled (1), file uploads skip the extension whitelist checks, as expected.
- Retains existing validation when the setting is disabled, ensuring secure defaults.

### Testing
- Verified that enabling “Upload any extension file” in the UI allows uploading files with arbitrary extensions.
- Confirmed that disabling the setting restores the original extension restrictions.

### Why
- Resolves an issue where uploading files with any extension was not functioning as intended, despite the configuration.

### DIFF
--- a/sources/upload.attachments.php
+++ b/sources/upload.attachments.php
@@ -261,17 +261,21 @@ if (strlen($file_name) == 0 || strlen($file_name) > $MAX_FILENAME_LENGTH) {
 
 // Validate file extension
 $ext = strtolower(getFileExtension($_REQUEST['name']));
-if (
-    in_array(
-        $ext,
-        explode(
-            ',',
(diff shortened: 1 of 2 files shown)


nilsteampassnet:
THank you @gudmmk 

Merged manually after local testing via develop

PR #4804, opened 2025-07-10
Fix: language strings for server internet connectivity status

gudmmk:
### Summary
Corrects the `server_not_connected_to_internet` string mistakenly said  "Server is connected to internet" instead of indicating it is **not** connected.

### Details
- Updated `server_not_connected_to_internet` in several language files to correctly read "Server is NOT connected to the internet".
- Helps avoid misleading messages on the admin Information page.

### Notes
- Minor language-only fix, no functional changes.


nilsteampassnet:
@gudmmk 
Thank you for the finding.
Nevertheless, this PR will not be merged as is.
All strings are managed through POEditor.
Changes need to be done there.
I will perform the proposed change.

PR #4891, opened 2025-10-23
Fix abnormal display of time zone and time format

FengZh61:
Fixed abnormal display of time zone and time format


nilsteampassnet:
Thank you

PR #4931, opened 2025-11-25
Fix HTML entities display in logs page

guerricv:
### Summary

This PR fixes the display of HTML entities in the logs page (Journaux):

- User names and labels were shown as `M&eacute;lanie`, `F&amp;H`, etc.
- We now decode HTML entities on the client side for the DataTables used in:
  - Connexions
  - Connexions échouées
  - Erreurs
  - Copie faite
  - Administration
  - Éléments

### Technical details

- Add a `decodeHtmlEntities()` helper in `pages/utilities.logs.js.php`.
- Use DataTables `columnDefs.render` for the columns containing user names / labels so they are decoded only for display, without touching the database values.

Tested on my instance (3.1.5.1) with French locale. Should also work on 3.1.5.3 as the file hasn't been touched between those releases.

### DIFF
--- a/pages/utilities.logs.js.php
+++ b/pages/utilities.logs.js.php
@@ -90,6 +90,16 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
     var oTableAdmin;
     var oTableErrors;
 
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
Thank you @guerricv 

Merged localy until now.

PR #4932, opened 2025-11-25
Decode HTML entities in email subject

guerricv:
Subject line is HTML-encoded (&apos;)

In French (and probably other languages), the subject currently appears like: Image

The translations store apostrophes as HTML entities (&apos;, etc.), and the subject is used as-is when creating the background task, so the mail client shows the literal entity in the subject.

On my instance I fixed this by decoding HTML entities in sources/main.functions.php, function prepareSendingEmail(), before storing the subject in background_tasks.

### DIFF
--- a/sources/main.functions.php
+++ b/sources/main.functions.php
@@ -1051,7 +1051,7 @@ function prepareSendingEmail(
             'created_at' => time(),
             'process_type' => 'send_email',
             'arguments' => json_encode([
-                'subject' => $subject,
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
@guerricv 
Sorry I already fixed the issue prior to see this PR.
The fix is now in 3.1.5.4

guerricv:
@nilsteampassnet No problem :) Well done !

PR #4935, opened 2025-11-26
Fix admin dashboard json responses and login type error on null session url - 3.1.5.4

guerricv:
### Summary

This PR fixes two issues observed on Teampass 3.1.5.4:

1. **Admin dashboard AJAX errors**: the browser console shows repeated  
   `AJAX Error: parsererror SyntaxError: Unexpected token 'e', "eyJ..." is not valid JSON`  
   for requests to `sources/admin.queries.php`.

2. **500 error on login**: `sources/identify.php` returns HTTP 500 with a PHP fatal error  
   when `buildAuthResponse()` is called with a `null` session URL.

---

#### Problem

On the admin dashboard, several AJAX calls are made with `dataType: 'json'` in JavaScript.
However, for the following `type` values handled by `sources/admin.queries.php`:

- `get_dashboard_stats`
- `get_live_activity`
- `get_system_status`
- `get_system_health`

the PHP code returns the response using:

```php
echo prepareExchangedData($data, 'encode');

### DIFF
--- a/sources/admin.queries.php
+++ b/sources/admin.queries.php
@@ -2777,6 +2777,543 @@ switch ($post_type) {
         );
 
         break;
+        
+    case 'get_dashboard_stats':
+    /**
+     * Get dashboard statistics
+     * Returns: users, items, folders, and logs statistics
+     * 
+     * @return array {
+     *   users: {active: int, online: int, blocked: int},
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
@guerricv 
Thank you
Fix another way.

Issue #4947, opened 2025-12-02
[3.1.5.8] - First OTP code for a new user is rejected


guerricv:
<h2>First of all: Nills, thank you for the huge work on 3.1.5.x</h2>
<p>
A big thank you for all the work done on the 3.1.5.x branch.
The improvements and bug fixes are really appreciated on our side – we can clearly see the amount of effort invested in this branch 👏
</p>

<h2>Environment</h2>
<ul>
  <li>TeamPass version: <strong>3.1.5.8</strong></li>
  <li>Previous version: <strong>3.1.5.1</strong> (upgrade path: 3.1.5.1 → 3.1.5.8)</li>
  <li>Web server: Apache 2 + PHP-FPM</li>
  <li>PHP version: 8.4</li>
  <li>Database: MySQL/MariaDB</li>
  <li>Instance: test environment restored from a recent backup of our production instance</li>
</ul>

<h2> First OTP code for a new user is rejected</h2>

<h3>Observed behaviour</h3>
<p>
When creating a brand new user from the admin interface, the first OTP/secret code generated for this user is not accepted at first login.
The user gets the following error message (French UI):
</p>
<pre>Impossible de décoder, merci de vérifier le code secret saisi.</pre>
<p>
The workaround is:
</p>
<ul>
  <li>The user triggers an <strong>OTP reset</strong> on first login.</li>
  <li>A new OTP/secret code is generated.</li>
  <li>The login works correctly and the user can access TeamPass.</li>
</ul>

<h3>Steps to reproduce</h3>
<ol>
  <li>Log in as an administrator.</li>
  <li>Create a new user account from the admin interface.</li>
  <li>Note the OTP/secret code displayed for this new user.</li>
  <li>Ask the user to log in and enter this code at first login.</li>
  <li>The user sees the red error message: <code>Impossible de décoder, merci de vérifier le code secret saisi.</code></li>
  <li>The user then uses the <strong>Reset OTP</strong> function.</li>
  <li>A new tranparant OTP code is generated; appication access works fine.</li>
</ol>

<h3>Expected behaviour</h3>
<ul>
  <li>The very first OTP/secret code generated for a new user should be valid and accepted at first login.</li>
  <li>The user should not need to reset OTP on their very first connection.</li>
</ul>

<h3>Additional notes</h3>
<ul>
  <li>I could only test the OTP displayed in the admin interface. I assume the email OTP is the same value, but I have not been able to test the email flow yet.</li>
[...]

PR #4948, opened 2025-12-03
Hotfix/fix 4947

nilsteampassnet:
Fix for #4947


guerricv:
@nilsteampassnet with this adjusted identify.php a got a double notification on users' connection.

nilsteampassnet:
Should be fixed with new commit

PR #4986, opened 2025-12-16
Fix DB Setup Step 5

fthomys:
Fixes "Query error occurred: Incorrect index name 'unique_item_id '"


nilsteampassnet:
Merged, thanks

PR #4988, opened 2025-12-17
Fix syntax error in deleting index

FengZh61:
`DROP INDEX IF EXISTS` is not supported in mysql and Mariadb, resulting in the upgrade process flushing index not successfully deleted

### DIFF
--- a/install/upgrade_run_3.1.5.php
+++ b/install/upgrade_run_3.1.5.php
@@ -138,7 +138,26 @@ foreach ($columns_to_add as $column_info) {
 // Drop previous index
 $tables = ['sharekeys_items', 'sharekeys_fields', 'sharekeys_files', 'sharekeys_suggestions', 'sharekeys_logs'];
 foreach ($tables as $table) {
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
@FengZh61 
Thank you but this SQL instruction is valid for MariaDB since years.
INdeed for MySQL, it is not the case.

I would suggest this approach to limit the loop effect:
```
    $check_index = mysqli_query($db_link, "
        SELECT INDEX_NAME 
        FROM information_schema.STATISTICS 
        WHERE TABLE_SCHEMA = DATABASE() 
        AND TABLE_NAME = `{$pre}{$table}`
        AND INDEX_NAME = `idx_object_user`
        LIMIT 1
    ");

    // If index exists then drop it
    if (mysqli_num_rows($check_index) > 0) {
        mysqli_query($db_link, "DROP INDEX `idx_object_user` ON `{$pre}{$table}`");
    }
```

OK for you?

nilsteampassnet:
Merged

PR #5131, opened 2026-03-12
fix table user missing  private_key_xss_migration column

FengZh61:
The private_key_xss_migration field is missing when creating the user table in the new installation.


nilsteampassnet:
Thank you
It has been merged

Issue #5139, opened 2026-03-24
Deleting an element from element list will show all duplicated elements


no-coders:
Hello,
trying to remove an element from elements list will immediately show all the other elements duplicated in the list.

They will disappear reloading page.

In the example test I created 3 element: test1, test2, test3. Deleting test1 for example the other are shown duplicated.

<img width="932" height="897" alt="Image" src="https://github.com/user-attachments/assets/dead02dc-05da-46b4-b919-ec2fdded7002" />

<img width="938" height="890" alt="Image" src="https://github.com/user-attachments/assets/c989f393-4f3e-4aa1-8d21-432fa69b0031" />

Have a nice day

guerricv:
Hello @no-coders 
Thanks for your feedback on that release, it helps !

I fixed that point in https://github.com/nilsteampassnet/TeamPass/pull/5141 and also your first issue https://github.com/nilsteampassnet/TeamPass/issues/5138 - Let's give it a try and give us your feed back :)

nilsteampassnet:
Fixed in 3.1.7.2

Issue #5138, opened 2026-03-24
Existing password complexity recognition will not work if editing an element using pencil icon


no-coders:
Hello,
actually I am using version 3.1.7.1. Seems that if I would like to edit an element (not folder element), if I try to access editing mode directly from "pencil icon" in the item list system is not able to recognize password complexity and If I would like to change any parameter except for password and try to save it I will always recive error that password complexity is not enough.

<img width="1537" height="542" alt="Image" src="https://github.com/user-attachments/assets/2257f48c-bbac-4f78-9164-fbc53abb2fcd" />

To bypass problem I have to edit password field, for example adding a char and removing it, and at that point password complexity is correctly recognized.

Just low priority problem :)

nilsteampassnet:
Fixed in 3.1.7.2

PR #5141, opened 2026-03-25
Fix item password complexity refresh/generation consistency and list duplicate refresh after deletion

guerricv:
<h2>Summary</h2>
<p>
This PR fixes several UI/UX inconsistencies in the item edition flow related to password complexity,
and also resolves a refresh issue in list view after deleting an item.
</p>

<h2>Issues addressed</h2>
<ul>
  <li>
    <strong>Password complexity bar inconsistency when editing an item</strong><br>
    The displayed password strength was not always correct depending on how the user entered edit mode:
    <ul>
      <li>directly from the list view using the pencil icon,</li>
      <li>or from the item detail view and then clicking edit.</li>
    </ul>
    In practice, the strength bar and hidden complexity value were not always recalculated the same way.
  </li>
  <li>
    <strong>Password generation not aligned with the folder-required complexity</strong><br>
    Clicking the generate button could return a password that did not match the complexity level required by the folder.
    The generation flow is now aligned with the requested folder complexity level.
  </li>
  <li>
    <strong>Duplicate items displayed in list view after deletion</strong><br>
    Deleting an item from list view could temporarily duplicate the remaining rows until the page was refreshed.
    This refresh sequence has been corrected.
  </li>
</ul>

<h2>Root causes</h2>
<ul>
  <li>
    The password strength plugin was not refreshed consistently in all edit entry points.
  </li>
  <li>
    The hidden field storing the current password complexity could remain outdated when editing an existing item.
  </li>
  <li>
    Password generation was not reliably targeting the folder complexity level shown in the UI.
  </li>
  <li>
    The list refresh logic after deletion could trigger an inconsistent redraw, leading to duplicated rows.
  </li>
</ul>

<h2>Changes made</h2>
<ul>
  <li>
    Unified password strength recalculation behavior for both edit flows:
    <ul>
      <li>edit from list view,</li>
      <li>edit from item detail view.</li>
    </ul>
  </li>
  <li>
    Ensured the password strength bar and the hidden complexity value are refreshed correctly
    when an existing item password is loaded into the edit form.
  </li>
  <li>
    Adjusted password generation so that it progressively reaches the folder-requested complexity level
    instead of stopping at an arbitrary lower level.
  </li>
  <li>
    Preserved a coherent UI update during generation so the final generated password matches the expected folder policy.
  </li>
  <li>
    Fixed the list refresh behavior after item deletion to prevent duplicate row rendering.
  </li>
</ul>

<h2>Files impacted</h2>
<ul>
  <li><code>items.js.php</code></li>
</ul>

<h2>Functional result</h2>
<ul>
  <li>
    The password complexity bar now behaves consistently regardless of how the user enters edit mode.
  </li>
  <li>
    The generated password now matches the requested folder complexity level much more reliably.
  </li>
  <li>
    Deleting an item from list view no longer causes duplicated items to appear.
  </li>
</ul>

<h2>Manual tests performed</h2>
<ul>
  <li>
    Opened an item directly from list view with the pencil icon and verified that the displayed complexity is correct.
  </li>
  <li>
    Opened an item from detail view, then clicked edit, and verified that the displayed complexity is also correct.
  </li>
  <li>
    Generated a new password in a folder with a required complexity level and verified that the resulting password reaches the expected level.
  </li>
  <li>
    Deleted an item from list view and verified that the remaining list is refreshed without duplicated rows.
  </li>
</ul>

<h2>Additional notes</h2>
<ul>
  <li>No database change.</li>
  <li>No language string added.</li>
  <li>No backend API change.</li>
</ul>

### DIFF
--- a/includes/language/english.php
+++ b/includes/language/english.php
@@ -2089,4 +2089,5 @@ return array(
     'network_security_auto_comment_current_ip' => 'Automatically added current admin IP',
     'network_security_auto_comment_server_ip' => 'Automatically added TeamPass server IP',
     'network_security_remote_addr' => 'Client remote address',
+    'password_generation_not_compliant_with_folder_complexity' => 'The selected password generation options do not allow reaching the folder required complexity. Please adjust the options or the length and try again.',
     );
--- a/includes/language/french.php
+++ b/includes/language/french.php
@@ -2097,5 +2097,5 @@ return array(
     'network_security_remote_addr' => 'Adresse distante du client',
     'email_subject_on_user_lock' => '[TeamPass] Un compte utilisateur a été bloqué',
     'email_body_on_user_lock' => 'Bonjour,<br><br>Ceci est un email généré par TeamPass.<br><br>Le compte utilisateur <b>#tp_user#</b> a été bloqué par la protection anti brute force le #tp_date# à #tp_time#.<br><br>Détails :<ul><li>Nom : #tp_name#</li><li>Email : #tp_email#</li><li>Adresse IP source : #tp_ip#</li><li>Déblocage automatique prévu : #tp_unlock_at#</li></ul><br>Cordialement.',
-
+    'password_generation_not_compliant_with_folder_complexity' => 'Les options de génération sélectionnées ne permettent pas d’atteindre la complexité requise par le dossier. Merci d’ajuster les options ou la longueur puis de réessayer.',
 );
--- a/pages/items.js.php
+++ b/pages/items.js.php
@@ -3087,45 +3087,6 @@ $var['hidden_asterisk'] = '<i class="fa-solid fa-asterisk mr-2"></i><i class="fa
         return Math.min(maximumSize, Math.max(minimumSize, configuredDefaultSize));
     }
 
-    function rememberBestPasswordCandidate(bestCandidate, candidate, requiredComplexity) {
-        if (!candidate) {
-            return bestCandidate;
-        }
-
-        const generatedComplexity = parseInt(candidate.generatedComplexity ?? 0, 10) || 0;
-        const meetsRequiredComplexity = generatedComplexity >= requiredComplexity;
-        const complexityDistance = meetsRequiredComplexity === true
-            ? generatedComplexity - requiredComplexity
-            : requiredComplexity - generatedComplexity;
-        const enrichedCandidate = $.extend({}, candidate, {
-            generatedComplexity: generatedComplexity,
-            meetsRequiredComplexity: meetsRequiredComplexity,
-            complexityDistance: complexityDistance,
-        });
-
-        if (!bestCandidate) {
-            return enrichedCandidate;
-        }
-
-        if (enrichedCandidate.meetsRequiredComplexity === true && bestCandidate.meetsRequiredComplexity !== true) {
-            return enrichedCandidate;
-        }
-
-        if (enrichedCandidate.meetsRequiredComplexity !== true && bestCandidate.meetsRequiredComplexity === true) {
-            return bestCandidate;
-        }
-
-        if (enrichedCandidate.complexityDistance < bestCandidate.complexityDistance) {
-            return enrichedCandidate;
-        }
-
-        if (enrichedCandidate.complexityDistance > bestCandidate.complexityDistance) {
-            return bestCandidate;
-        }
-
-        return bestCandidate;
-    }
-
     function generatePasswordRespectingFolderComplexity(elementId, passwordGenerationOptions, requiredComplexity, searchState = null) {
         const normalizedRequiredComplexity = parseInt(requiredComplexity, 10) || 0;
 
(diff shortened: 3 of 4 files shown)


no-coders:
Hello @guerricv,
new file, when fix #5141 is included in current version, solved #5139. Unfortunately seems to solve #5138 but generated 1 other bug:

1. If I edit item using "pencil", whatever parameters I am using to generate password pattern (es. length, symbols), after password generation, regreet to default parameters and I visually see password chars disappearing in realtime from the new generated password. Basically, I am not able to keep a new password that **has not default** password **parameters** (es. length).

Basically, with manual tests yu performed: "Generated a new password in a folder with a required complexity level and verified that the resulting password reaches the expected level.", seems to be working only because you probably tested with default paramenters and/or checked the complexity bar. 

If you introduce a new test that is: "increasing password length about 5 chars up to default length and verify system is able to generate the required password parameters" you should be able to see the problem.

guerricv:
Hi,
You mean that if you enter a password more secure than the security level required by folder parameter you can't save it ?
Less secure, you won't be able to save it, that is expected. I'm currently checking again that file.

no-coders:
> Hi, You mean that if you enter a password more secure than the security level required by folder parameter you can't save it ? Less secure, you won't be able to save it, that is expected. I'm currently checking again that file.

You can save it but you should not be able to generate a password more complex than default parameters. For default parameters I am refeering to that length value, in the admin global settings

image

guerricv:
ok got it ! Thanks a lot :) Ping you for a new try when it is fixed :)

guerricv:
<p>
Thanks a lot <strong>@no-coders</strong> for the detailed testing and for the very relevant feedback.
Your report helped identify an important regression in the password generation flow.
</p>

<p>
I have pushed an updated correction with the following adjustments:
</p>

<ul>
  <li>
    <strong>Password complexity bar refresh is now consistent in both edit entry points</strong><br>
    The password strength bar is now recalculated correctly whether the user enters edit mode:
    <ul>
      <li>directly from the list view using the pencil icon,</li>
      <li>or from the item detail view and then clicks edit.</li>
    </ul>
    This also keeps the hidden complexity value synchronized with the displayed strength.
  </li>

  <li>
    <strong>Password generation now respects the folder-required complexity without overwriting user options</strong><br>
    The generator now keeps the user-selected generation options unchanged
    (length, lowercase, uppercase, digits, symbols, secure mode).
    It no longer resets the visible generation parameters after clicking the generate button.
  </li>

  <li>
    <strong>Generation now progressively increases password length when needed</strong><br>
    The generation flow now starts from the currently selected length.
    If the generated password does not reach the folder-required complexity,
    the logic retries with the same selected options and progressively increases only the length,
    until a compliant password is found or the configured maximum password length is reached.
  </li>

  <li>
    <strong>The length selector is updated to the actually used length</strong><br>
    When the generator needs a longer password to satisfy the folder policy,
    the UI length selector is automatically updated to the final length that was required.
    This keeps the UI coherent for the next generation attempt.
  </li>

  <li>
    <strong>Admin default password length is now used as the proper initial baseline</strong><br>
    The item password generation UI now correctly relies on the configured default password length,
    instead of falling back to an unrelated hardcoded/runtime value.
  </li>

  <li>
    <strong>Duplicate rows after item deletion in list view are fixed</strong><br>
    Deleting an item from list view no longer causes the remaining items to appear duplicated
    until a manual refresh is performed.
  </li>
</ul>

<p>
So in summary, this update fixes:
</p>

<ul>
  <li>the inconsistent complexity bar refresh when editing an item,</li>
  <li>the password generation regression reported during testing,</li>
  <li>the improper reset of generation parameters,</li>
  <li>the need to manually adjust length when folder complexity is higher than the current selection,</li>
  <li>and the list duplication issue after item deletion.</li>
</ul>

<p>
Thanks again for the careful review and for spotting the edge case with non-default generation parameters.
That feedback was essential to make the behavior both correct and user-friendly.
</p>

no-coders:
Hello @guerricv,
now seems to work enough good. I didn't find any big inconsistency with expected behaviour. I added string 'password_generation_not_compliant_with_folder_complexity' even to 'itaian.php' that I am using.

Waiting for next release.

Thank you for you work. Love it :)

guerricv:
Thanks a lot !

While testing, I noticed that changing a password, without websocket enable, is not reloading correctly the items list page after saving. Trying to find a fix for that and it will be all for today !

Sorry, i'm a little lazy with languages strings 😂

guerricv:
<p>
I have pushed an additional fix for another issue identified during testing when WebSocket is not enabled/configured on the instance.
</p>

<p>
The problem occurred when editing an item and changing its password:
the save completed server-side, but the UI remained stuck on the loading screen because the front-end was waiting only for the <code>task_completed</code> WebSocket event before reloading the item details.
When WebSocket is unavailable, that event never arrives.
</p>

<p>
This update adds a safe fallback mechanism:
</p>

<ul>
  <li>the current WebSocket-based behavior is preserved when WebSocket is available,</li>
  <li>a timed front-end fallback now reloads the item details if the WebSocket refresh does not occur,</li>
  <li>the fallback is automatically cancelled when the normal WebSocket refresh is received in time,</li>
  <li>the same logic has been applied to both <code>update_item</code> and <code>new_item</code> flows.</li>
</ul>

<p>
As a result, when WebSocket is disabled or not configured, saving an item after changing its password no longer leaves the page stuck on the loading spinner.
</p>

nilsteampassnet:
Thank you

PR #5155, opened 2026-03-30
Fix split view layout after saving an item

guerricv:
<p>This PR fixes a split view layout issue introduced when saving an item from the edit form.</p>

<h3>Summary</h3>
<p>
When split view mode is enabled in the user profile, saving an item could leave the details panel rendered on the far left side of the page,
hiding the items list instead of restoring the expected 3-column layout.
</p>

<h3>Root cause</h3>
<p>
After save, the UI temporarily switches to a near full-screen item view by hiding the tree and items list containers.
When the item details view is then reloaded in split mode, the layout classes were restored, but the hidden state of the left columns was not.
As a result, the details container became the first visible column and appeared on the left.
</p>

<h3>Fix</h3>
<ul>
    <li>When split view mode is active, explicitly restore the visibility of the tree and items list containers before applying the split layout classes.</li>
    <li>Keep the current wide edit/create form behavior unchanged, as it provides a better editing experience than a narrow split layout.</li>
</ul>

<h3>Result</h3>
<p>
After saving an item in split view mode, the page now correctly returns to the expected layout:
tree on the left, items list in the middle, and item details on the right.
</p>

<h3>Testing</h3>
<ul>
    <li>Split view enabled</li>
    <li>Open an item in edit mode</li>
    <li>Save changes</li>
    <li>Confirmed that the details panel is restored on the right and the items list remains visible</li>
</ul>

<p>
Also includes a very small French translation correction.
</p>

### DIFF
--- a/includes/language/french.php
+++ b/includes/language/french.php
@@ -471,7 +471,7 @@ return array(
     'settings_upload_imageresize_options_w' => 'Redimensionnement des Images en Largeur (en pixels)',
     'settings_upload_imageresize_options_h' => 'Redimensionnement des Images en Hauteur (en pixels)',
     'settings_upload_imageresize_options_q' => 'Qualité de l’Image redimensionnée',
-    'settings_importing' => 'Permettre d’importer des données depuis des fichiers CVS/KeyPass',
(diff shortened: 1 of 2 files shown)


nilsteampassnet:
Thank you

PR #5169, opened 2026-04-01
Persist browser extension FQDN automatically when enabling API

guerricv:
<h2>Summary</h2>
<p>This PR fixes an inconsistency in the API configuration page related to the browser extension FQDN.</p>

<p>The browser extension tab could display a detected default FQDN derived from <code>cpassman_url</code>, but this value was only prefilled in the UI and was not always persisted in configuration. As a result, backend logic could still operate with an empty <code>browser_extension_fqdn</code> until the field was manually saved.</p>

<h2>What is changed</h2>
<ul>
  <li>Add a silent default persistence of <code>browser_extension_fqdn</code> when API is enabled and no value is already stored.</li>
  <li>Implement the behavior in <code>save_option_change</code>, which is the actual save flow used by the API settings page.</li>
  <li>Keep the same protection in <code>save_api_status</code> as a fallback for alternative or legacy entry points.</li>
  <li>Preserve any existing non-empty <code>browser_extension_fqdn</code> value and never overwrite it.</li>
</ul>

<h2>Why</h2>
<ul>
  <li>Keep frontend display and backend configuration aligned.</li>
  <li>Avoid requiring a manual save of the FQDN after enabling the API.</li>
  <li>Ensure browser extension related backend checks can rely on a persisted FQDN value immediately after activation.</li>
</ul>

<h2>Tests</h2>
<ul>
  <li>Tested with no existing <code>browser_extension_fqdn</code>: enabling API now creates and stores the detected value automatically.</li>
  <li>Tested with an existing non-empty <code>browser_extension_fqdn</code>: enabling API keeps the stored value unchanged.</li>
  <li>Confirmed the behavior is silent and does not introduce any UI regression.</li>
</ul>

### DIFF
--- a/sources/admin.queries.php
+++ b/sources/admin.queries.php
@@ -1800,8 +1800,9 @@ switch ($post_type) {
                         if ($path !== '') {
                             $segments = explode('/', $path);
                             if (
-                                $segments[0] !== ''
-                                && strpos($segments[0], '.php') === false
+                                isset($segments[0]) === true
+                                && $segments[0] !== ''
+                                && strpos((string) $segments[0], '.php') === false
                             ) {
                                 $host = (string) $segments[0];
                             }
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
Thank you

PR #5173, opened 2026-04-03
Remove redundant paragonie/random_compat dependency

brendt:
This package depends on `paragonie/random_compat: ^9.99`, but also `php: >=8.1`. Since `random_bytes` and `random_int` were added in PHP 7.0, the dependency on `paragonie/random_compat` seems redundant?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -185,7 +185,6 @@
         "phpmailer/phpmailer": "^6.8",
         "moxiecode/plupload": "^3.1",
         "tecnickcom/tcpdf": "^6.6",
-        "paragonie/random_compat": "^9.99",
         "elegantweb/sanitizer": "^2.1",
         "defuse/php-encryption": "^2.4",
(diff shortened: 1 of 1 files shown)


nilsteampassnet:
The dependency is indeed no longer directly used by TeamPass (since PHP >= 8.2 natively provides random_bytes() and random_int()).
However, the ^9.99 version pinned here is an empty meta-package published by Paragonie: its lib/random.php file contains nothing but a comment and // NOP.
It is retained because two transitive dependencies (defuse/php-encryption and phpseclib/phpseclib) still list paragonie/random_compat in their require block.
The ^9.99 constraint ensures that this empty version is installed instead of a 2.x variant with actual code.
While removing the line would likely work, it would leave Composer free to resolve the transitive constraints with another version. Therefore, it is prefered to maintain this explicit lock.

PR #5192, opened 2026-04-14
Add self-service pwd recovery for local accounts and Email localization fix for create_user_keys notifications

guerricv:
## Summary

This change introduces a self-service password recovery flow for local TeamPass accounts.

When a login attempt fails for an identified local user, the login page can now display an **"I lost my password"** action. If the user requests recovery, TeamPass generates a temporary password, sends it by email, and forces the user to change it at the next login.

The feature is now controlled by an administration setting exposed in the email configuration page, allowing administrators to enable or disable the recovery flow without code changes.

## User Flow

1. A user enters a valid local login with an incorrect password.
2. The authentication response indicates that password recovery is available for that account.
3. The login page shows an **"I lost my password"** action under the login button.
4. If the user clicks it, TeamPass generates a temporary password for that local account.
5. The temporary password is queued for delivery by email through the existing TeamPass email system.
6. The account is marked so that the user must change the password at the next successful login.
7. After logging in with the temporary password, the existing forced password change flow is triggered.

## Administrative Control

A new administration option has been added to the email configuration screen:

- **Setting name:** `enable_local_password_recovery`
- **Location:** Email configuration page
- **Purpose:** Enable or disable self-service password recovery for local accounts
- **Default for new installs:** Enabled

For backward compatibility, the new setting is synchronized with the historical `disable_show_forgot_pwd_link` flag so existing installations keep a coherent behavior.

### 1. Login UI behavior

- The login page now contains a hidden recovery action below the standard login button.
- The action stays hidden by default and only becomes visible after a failed login for an eligible local account.
- The action is automatically hidden again when the user edits the login or password fields, or when a new authentication attempt starts.

### 2. Eligibility rules

Password recovery is only offered when all of the following conditions are met:

- The account exists and uses `auth_type = local`
- The feature is enabled in administration
- Email sending is configured
- The user has a non-empty email address
- The user is not disabled
- No key-generation or similar background process is already ongoing for the user

### 3. Request authorization guard

The backend does not accept a recovery request blindly.

The request is tied to a short-lived session context created immediately after a failed login attempt for the same account. This means the recovery action must come from the same browser session and shortly after the failed login, which reduces abuse and keeps the flow aligned with the UI.

### 4. Temporary password generation

When recovery is requested, TeamPass generates a temporary password and reuses the existing key regeneration flow already used for password resets.

The implementation integrates with the current `handleUserKeys()` and background task pipeline instead of introducing a parallel password reset mechanism.

### 5. Forced password change at next login

Once the temporary password has been generated, the user state is finalized with the existing `auth-pwd-change` special status.

This reuses the current forced password change UX and avoids introducing another special-case screen.

### 6. Email delivery

The temporary password is delivered through the TeamPass email queue using the existing email infrastructure.

A dedicated email subject and body were added in English and French for this flow.

### 7. Failed-attempt cleanup

After a successful recovery request, login-related failure entries are cleared for the account and the session password attempt counter is reset.

This prevents the newly recovered user from being blocked by stale failed-authentication state.

## Implementation Notes

| Area | Files | Details |
|---|---|---|
| Login UI | `includes/core/login.php`, `includes/core/login.js.php` | Adds the recovery action, shows it only after eligible failed logins, and posts the recovery request. |
| Authentication backend | `sources/identify.php` | Exposes the recovery endpoint, evaluates eligibility, protects the request with session context, triggers temporary password generation, resets failed-attempt state, and queues the email. |
| Password/key regeneration | `sources/main.functions.php`, `scripts/traits/UserHandlerTrait.php` | Extends the existing key-generation pipeline so the final user special state can be explicitly set to `auth-pwd-change`. |
| Administration | `pages/emails.php`, `sources/admin.queries.php`, `pages/ldap.php` | Moves the control to the email settings page, saves the new option, and keeps the legacy flag synchronized for compatibility. |
| Installation defaults | `install/install-steps/run.step5.php` | Enables the new setting by default for fresh installations. |
| Translations | `includes/language/english.php`, `includes/language/french.php` | Adds the recovery CTA, success/error messages, email subject/body, and administration labels. |

## Why This Approach

- It reuses existing TeamPass password and key management flows instead of duplicating sensitive logic.
- It keeps the recovery experience limited to local accounts, which matches the actual password ownership model.
- It fits into the current forced-password-change behavior already supported by the application.
- It introduces an administrator-controlled switch so the feature can be turned on or off depending on policy.
- It remains backward compatible with the historical forgot-password visibility flag.

## Security Considerations

- The feature only applies to local accounts.
- The recovery action is not shown unless a login failure occurred for an eligible account.
- The backend requires a short-lived session-based authorization context tied to the failed login before accepting the recovery request.
- The account is forced through a password change on the next login.
- The flow is disabled when email delivery is not configured or when the account is not eligible.

## Backward Compatibility

- Existing instances that only have `disable_show_forgot_pwd_link` continue to behave correctly because the new logic falls back to that value when the new setting does not exist yet.
- When an administrator saves the new option, the legacy flag is updated in parallel so old references remain coherent.
- No dedicated database migration is required for existing systems because the setting is stored through the standard admin option persistence flow.

## Validation

Manual validation covered the main expected behavior:

- Failed login for a local account shows the recovery action
- Submitting the recovery action generates a temporary password flow and returns a success message
- The account is flagged for password change on next login
- The administrator can now enable or disable the feature from the email settings page
- Legacy installations remain compatible through the fallback and synchronization logic

## Suggested Reviewer Focus

- Eligibility rules for local account recovery
- Session-based authorization of the recovery request
- Reuse of the existing key-generation and forced-password-change pipeline
- Admin setting placement and backward compatibility with the legacy flag
- Email queue integration and user-facing messages


guerricv:
## Email localization fix for `create_user_keys` notifications

This PR fixes the language selection used for emails sent at the end of the `create_user_keys` background task.

### Root cause

Two issues were combined here:

1. In `scripts/traits/UserHandlerTrait.php`, the background worker was instantiating `Language('english')` directly when sending the final notification email.
2. Some call sites were passing an already translated email body instead of a translation key, which meant the body language could depend on the session that triggered the action rather than on the target user.

As a result, users could receive:
- a fully English email, or
- an English subject with a body translated in the admin/session language.

### What was changed

#### 1. Use the target user's language in the worker
In `scripts/traits/UserHandlerTrait.php`, the final notification email now resolves its language from the destination user's `user_language`, with fallback to the instance `default_language` and then to `english`.

This ensures that both:
- the subject (`login_credentials`)
- and the body (`email_body_user_config_*`)

are translated using the recipient's language.

#### 2. Pass translation keys instead of pre-translated strings
In:
- `sources/identify.php`
- `sources/users.queries.php`

the flows that were still passing `$lang->get(...)` now pass the translation key itself instead, for example:
- `email_body_user_config_2`
- `email_body_user_config_6`
- `email_body_user_config_4`

This keeps translation resolution centralized in the background worker and avoids language drift between the triggering session and the final recipient.

### Impacted flows

This adjustment covers the email notifications generated by `create_user_keys`, including:

- admin-triggered AD/LDAP user key regeneration
- admin-triggered local password regeneration/change
- account-ready notifications after key generation
- automatic LDAP first-login account creation
- automatic OAuth2 first-login account creation
- manual local user creation flows relying on the same background task

### Important note about `default_language`

The behavior where newly created LDAP/OAuth2 users inherit `default_language` is unchanged and is still expected.

That setting is the instance default language used for:
- new accounts
- the login page

This fix only changes how the final email language is selected when sending `create_user_keys` notifications.

### Result

After this change, the email subject and body are consistently translated using the destination user's language instead of being forced to English or tied to the triggering session language.

PR #5198, opened 2026-04-20
fix(files): strip PKCS#7 padding when decrypting item attachments

Valentin261-1101:
encryptFile() uses phpseclib AES-CBC with default PKCS#7 padding.
decryptFile() called disablePadding(), so padding bytes were left in
the decrypted plaintext (e.g. three 0x03 bytes when file length ≡ 13 mod 16).

Remove disablePadding() so decryption matches encryption and attachment
downloads/previews match the original file bytes.


nilsteampassnet:
Merged.

THank you

PR #5199, opened 2026-04-21
Restore and improve Knowledge Base, add local password recovery, fix create_user_keys email localization, and preserve HTML-sensitive password integrity

guerricv:
## Summary

This PR consolidates several related improvements and fixes that were originally prepared separately.

It includes:

- a new self-service password recovery flow for local TeamPass accounts
- a fix for language selection in `create_user_keys` notification emails
- the restoration of the Knowledge Base feature in TeamPass v3
- significant functional and UI improvements for the Knowledge Base
- Knowledge Base activity integration into the admin live activity feed
- a fix for password integrity issues involving HTML-sensitive sequences such as `<` and literal entities like `&lt;`

Taken together, these changes improve user autonomy, consistency of user-facing communications, restoration of a partially missing feature, and data integrity across password handling flows.

---

## 1. Local self-service password recovery for local accounts

This PR introduces a self-service password recovery flow for local TeamPass accounts.

When a login attempt fails for an identified local user, the login page can now expose an **"I lost my password"** action. If the user requests recovery, TeamPass generates a temporary password, sends it through the existing email system, and forces the user to change it at the next successful login.

### User flow

1. A user enters a valid local login with an incorrect password.
2. The authentication response indicates whether recovery is available for that account.
3. The login page displays an **"I lost my password"** action under the login button.
4. If the user clicks it, TeamPass generates a temporary password for that local account.
5. The temporary password is queued for delivery through the existing TeamPass email queue.
6. The account is flagged so the user must change the password at the next login.
7. After login with the temporary password, the existing forced password change flow is triggered.

### Administrative control

A new administration option has been added to the email configuration page:

- **Setting name:** `enable_local_password_recovery`
- **Location:** Email configuration page
- **Purpose:** Enable or disable self-service password recovery for local accounts
- **Default for new installs:** Enabled

For backward compatibility, this new setting is synchronized with the historical `disable_show_forgot_pwd_link` flag so existing installations keep coherent behavior.

#### Login UI behavior

- The login page now contains a hidden recovery action below the standard login button.
- The action stays hidden by default and only becomes visible after a failed login for an eligible local account.
- The action is hidden again when the user edits the login or password fields, or when a new authentication attempt starts.

#### Eligibility rules

Password recovery is offered only when all of the following conditions are met:

- the account exists and uses `auth_type = local`
- the feature is enabled in administration
- email sending is configured
- the user has a non-empty email address
- the user is not disabled
- no key-generation or similar background process is already ongoing for the user

#### Request authorization guard

The backend does not accept a recovery request blindly.

The request is tied to a short-lived session context created immediately after a failed login attempt for the same account. This means the recovery action must come from the same browser session and shortly after the failed login, which reduces abuse and keeps the flow aligned with the UI.

#### Password generation and account state

The recovery flow reuses the existing key regeneration pipeline instead of introducing a separate reset implementation.

Once the temporary password has been generated, the user is finalized with the existing `auth-pwd-change` special status so the already supported forced-password-change UX is reused as-is.

#### Email delivery

The temporary password is delivered through the existing TeamPass email queue.

Dedicated email subject/body translations were added for English and French.

#### Failed-attempt cleanup

After a successful recovery request:

- login-related failed-authentication entries are cleared for the account
- the session password attempt counter is reset

This prevents the recovered user from being blocked by stale failure state.

### Why this approach

- It reuses existing TeamPass password/key handling flows instead of duplicating sensitive logic.
- It limits the recovery feature to local accounts, which matches the actual password ownership model.
- It fits into the current forced password change behavior already supported by the application.
- It is administrator-controlled.
- It remains backward compatible with the historical forgot-password visibility flag.

### Security considerations

- Applies only to local accounts
- Recovery action is shown only after a failed login for an eligible account
- Backend requires a short-lived session-based authorization context tied to the failed login
- User is forced to change password at next login
- Flow is disabled when email delivery is not configured or the account is not eligible

---

## 2. Fix email localization for `create_user_keys` notifications

This PR fixes the language selection used for emails sent at the end of the `create_user_keys` background task.

### Root cause

Two issues were involved:

1. `scripts/traits/UserHandlerTrait.php` was instantiating `Language('english')` directly when sending the final notification email.
2. Some call sites were passing an already translated email body instead of a translation key, which meant the body language could depend on the triggering session instead of the target user.

As a result, users could receive:

- a fully English email, or
- an English subject with a body translated in the admin/session language

#### Use the target user's language in the worker

The final notification email now resolves its language from the destination user's `user_language`, with fallback to the instance `default_language`, then to English.

This ensures that both:

- the subject (`login_credentials`)
- the body (`email_body_user_config_*`)

are translated using the recipient's language.

#### Pass translation keys instead of pre-translated strings

The affected flows now pass translation keys instead of `$lang->get(...)` values.

Examples:

- `email_body_user_config_2`
- `email_body_user_config_6`
- `email_body_user_config_4`

This keeps translation resolution centralized in the background worker and avoids language drift between the triggering session and the final recipient.

### Impacted flows

This adjustment covers the email notifications generated by `create_user_keys`, including:

- admin-triggered AD/LDAP user key regeneration
- admin-triggered local password regeneration/change
- account-ready notifications after key generation
- automatic LDAP first-login account creation
- automatic OAuth2 first-login account creation
- manual local user creation flows relying on the same background task

### Important note

The behavior where newly created LDAP/OAuth2 users inherit `default_language` is unchanged and remains expected.

This fix only changes how the final email language is selected when sending `create_user_keys` notifications.

### Result

After this change, both email subject and body are consistently translated using the destination user's language instead of being forced to English or tied to the triggering session language.

---

## 3. Restore the Knowledge Base feature in TeamPass v3

This PR restores the Knowledge Base feature in TeamPass v3.

The feature was still partially present in the codebase:

- the `enable_kb` setting still existed
- item detail code could still load linked KB entries
- item UI code still expected KB badges/links

However, the dedicated page, related backend handlers, and menu integration were no longer available in v3.

Instead of copying the old v2 implementation as-is, this work rebuilds the feature so it matches the current v3 structure and UI patterns.

### Problem addressed

- The Knowledge Base option was still exposed in administration, but enabling it had no usable effect for end users.
- The menu entry was no longer effectively available in the v3 navigation flow.
- KB links could still be generated from items, but there was no complete user-facing KB feature behind them.
- This created a functional inconsistency and exposed a legacy feature that was only partially wired.

### Restored capabilities

- Knowledge Base menu entry for regular users when `enable_kb = 1`
- KB article listing
- KB article creation / edition / deletion
- KB category management
- direct access through `?page=kb&id=...`
- linking KB articles to TeamPass items
- display of linked KB entries on item pages
- attachment support for KB articles
- attachment validation using the same allowed extensions settings as the uploads configuration

### Behavioral choices

The KB is intentionally hidden from administrators in the main menu, because the current admin navigation flow does not rely on the standard items page and the feature is intended for end-user usage.

### Soft delete and logs

- KB article deletion now uses a soft delete approach
- deleted KB articles can be restored or permanently deleted from the recycle bin flow
- KB actions are logged: creation, display, modification, deletion, restore, purge
- KB activity is visible through logs/recycle-bin integrations

### Technical direction

This is not a raw copy of the historical v2 Knowledge Base code.

It is a v3-compatible rebuild based on:

- current page/backend organization
- current navigation/page loading flow
- existing TeamPass settings for visibility and upload restrictions
- language strings instead of hardcoded UI text

---

## 4. Knowledge Base functional and UI improvements

Beyond the restoration itself, this PR significantly improves the Knowledge Base experience.

### Better article rendering

- improved article readability
- preserved multi-line formatting when displaying articles and comments
- richer rendering for plain text content:
  - paragraph separation
  - line breaks
  - simple unordered lists
  - simple ordered lists
  - blockquotes
- clickable URLs preserved
- emoji support ensured

### Comment system for articles

A new comment system was added.

- each article now has an `allow_comments` option
- comments are disabled by default on new articles
- when enabled:
  - users who can view the article can post comments
  - comment authors can delete their own comments
  - article authors can delete inappropriate comments

No moderation workflow was added by design.

### Improved comment display

- comment author is displayed as **First name Last name**
- date and time are shown on the same line as the author for a cleaner layout
- line breaks are preserved properly in comments

### UI refresh

- cleaner and more modern article reading area
- improved spacing and card styling
- improved readability of article and comment sections
- adjusted viewer header actions alignment
- improved article list layout and column sizing
- improved visibility of action buttons on selected article header
- removed unstable visual jump/scroll behavior when opening an article

### Database changes

- added `allow_comments` field to KB entries
- added dedicated `kb_comments` table
- moved the KB upgrade logic into `upgrade_run_3.1.7.php`
- added charset/collation updates so KB content and comments support full `utf8mb4`, including emojis

### Important fixes included in the KB work

- fixed article/comment formatting being flattened during save/display
- fixed AJAX response decoding so rendered HTML is no longer stripped before display
- fixed item association persistence when editing an article:
  - linked items no longer visible to the current editor are no longer removed unintentionally
  - only visible/editable associations are modified
  - hidden existing links are preserved
- fixed KB attachment deletion flow consistency
- improved attachment workflow for new articles by handling pending uploads after initial save

### Result

The Knowledge Base now provides:

- cleaner article reading
- preserved formatting
- emoji-safe content
- optional comments per article
- safer persistence of linked items
- a more polished overall user experience

---

## 5. Include KB activity in the admin live activity feed

This PR also includes the related admin-side follow-up so Knowledge Base activity appears correctly in the live dashboard feed.

### Included updates

- `sources/admin.queries.php`: include KB events in live activity feed
- `pages/admin.js.php`: show KB activity actions correctly
- `pages/admin.php`: simplify live activity header styling

This keeps the dashboard consistent with the restored KB activity logging.

---

## 6. Preserve password integrity for HTML-sensitive values

This PR fixes a long-standing issue where passwords containing HTML-sensitive sequences such as `<` or the literal string `&lt;` could be altered, misread, or falsely reported as corrupted in some TeamPass flows.

The first visible symptom was a false positive in `scan_corrupted_items.php`, but the investigation showed a broader consistency issue across password write/read/display paths, including password history.

#### Passwords were modified before encryption

The secret encryption path was sanitizing password data before storage.

This created a mismatch between:

- the raw password length stored in `items.pw_len`
- the actual value encrypted and later decrypted

This could produce:

- false `len_mismatch` corruption reports
- inconsistent server-side reads
- different behavior depending on whether the value was old or newly saved

#### Some request/display paths still decoded literal HTML entities too early

Even after fixing secret storage, some paths could still transform a literal password such as `&lt;` into `<`.

Affected patterns included:

- item create/update decoding `pw` with `htmlspecialchars_decode()`
- non-encrypted request payload decoding over-decoding JSON content
- password history data being re-purified on the client before rendering

#### Secrets are now stored exactly as entered

The encryption path no longer sanitizes password content before encryption.

This is intentional: passwords are secrets, not HTML fragments, and must be preserved verbatim.

#### Legacy compatibility is preserved

A conservative legacy normalization step was introduced for historically entity-encoded passwords.

It only decodes when the decoded value length matches the stored `pw_len`, which avoids altering legitimate new passwords such as `&lt;`.

#### Non-encrypted JSON payload decoding is now conservative

The server now tries:

1. the raw JSON payload
2. one HTML entity decode
3. a second decode only if JSON is still invalid

This preserves literal entity sequences inside string values while still supporting HTML-escaped payload envelopes.

#### Item create/update no longer decodes `pw`

The item save path now keeps `pw` exactly as received instead of decoding it before encryption.

#### Current password display preserves literal entities

The display path was adjusted so the front end no longer reinterprets a literal returned value such as `&lt;` into `<`.

#### Password history also preserves literal entity values

Password history entries now use a password-specific key already excluded from generic client-side purification.

This prevents the history modal from converting a legitimate literal `&lt;` back into `<` during rendering.

#### Password-related server-side reads were aligned

The same normalization logic is now used consistently across key flows, including:

- item display
- edit form loading
- clipboard/copy flow
- export flows
- OTV-related reads
- password length / complexity recalculation
- corrupted-items scan

### Why this is safe

This PR does **not** perform any destructive migration and does **not** bulk-rewrite stored passwords.

Safety points:

- no schema change
- no mass re-encryption job
- no unconditional `html_entity_decode()` on all passwords
- legacy decoding only occurs when `pw_len` proves the decoded value is the intended original one
- output-side protection remains the correct place for XSS prevention

This improves data integrity without weakening security.

### Important limitation

If an item or password-history entry was already saved incorrectly during the buggy behavior window, it cannot always be reconstructed automatically.

For example, if a literal `&lt;` was already persisted as `<`, the original intent has already been lost in stored data.

This PR fixes future reads/writes and prevents new corruption or false interpretation, but it does not try to guess and rewrite already-lost historical values.

---

## Backward compatibility and upgrade notes

- Existing installations remain compatible with the historical forgot-password visibility flag through fallback and synchronization logic.
- No dedicated database migration is required for the password recovery setting because it uses the standard admin option persistence flow.
- The `create_user_keys` email localization change preserves the existing `default_language` behavior for new LDAP/OAuth2 users.
- Knowledge Base restoration follows the TeamPass v3 architecture instead of reviving legacy v2 patterns directly.
- KB schema changes are integrated into the existing upgrade flow.
- Password integrity fixes avoid destructive migration and keep compatibility with legacy data where safe to do so.

---

### Local password recovery

- failed login for a local account shows the recovery action
- submitting the recovery action generates the temporary password flow and returns a success message
- the account is flagged for password change on next login
- administrator can enable/disable the feature from the email settings page
- legacy installations remain compatible through fallback and synchronization logic

### Knowledge Base

- enable / disable Knowledge Base setting
- KB menu visibility for non-admin users
- KB article creation
- KB article edition
- KB article soft deletion and recycle bin flow
- KB logs visibility
- KB article linking to items
- KB links display on item page
- KB attachment upload flow
- upload restriction enforcement based on TeamPass upload settings

### Password integrity

- new password containing `<`
- new password containing literal `&lt;`
- display / reveal in item view
- clipboard copy
- loading password into edit form
- save without changing the password
- export flows
- corrupted-items scan no longer reporting false `len_mismatch`
- password history rendering for literal entity values
- regression check ensuring a real `<` still behaves as `<`

---

## Reviewer focus

Suggested areas to review:

- eligibility and session-based authorization for local password recovery
- reuse of the existing key-generation and forced-password-change pipeline
- admin setting placement and backward compatibility with the legacy forgot-password flag
- recipient-language resolution in `create_user_keys` email notifications
- Knowledge Base v3 reintegration and consistency with current navigation/page architecture
- KB attachment restrictions, logging, recycle-bin flow, and comment behavior
- password write/read/display consistency for HTML-sensitive values
- conservative legacy normalization and non-destructive compatibility behavior


guerricv:
## Additional Adjustment

During validation, we identified that the remaining issue was not only tied to the password storage/read logic itself, but also to the transport layer when `encryptClientServer` is disabled.

## Root Cause

When client/server exchange encryption is disabled, the item password is sent inside a plain JSON payload.

In that mode, the password can still be altered during request handling before the normal item encryption flow runs.  
This is why the behavior differed between instances:

- instance with `encryptClientServer = 1`: `<` was preserved correctly
- instance with `encryptClientServer = 0`: `<` could be transformed into `&lt;` during transport

So the previous fix was correct for storage and legacy normalization, but it still needed one extra adjustment to make the transport path reliable in both modes.

## What Was Changed

A targeted transport-level fix was added for item passwords:

- the password is now base64-encoded on the client before being sent in the item save payload
- a small flag is sent alongside it to indicate transport encoding
- the server decodes that transport value before applying the normal password handling and encryption flow

This approach was intentionally kept narrow:

- only the item password transport was adjusted
- the rest of the payload handling remains unchanged
- existing server-side validation and sanitization for the other fields are preserved
- backward compatibility is kept if an older client sends `pw` without the new transport flag

## Why This Approach

This is a safer and more focused fix than loosening sanitization on the whole JSON payload.

It ensures that password content remains byte-stable during transport even when client/server encryption is disabled, while avoiding broader side effects on the rest of the request processing.

## Impact

After this adjustment, password handling is now consistent whether `encryptClientServer` is enabled or disabled.

The following values are preserved correctly during item create/update flows:

- `<`
- `>`
- literal entity strings such as `&lt;`

## Files Updated

- `pages/items.js.php`
- `sources/items.queries.php`
- `sources/main.functions.php`
- `tests/Stubs/auth_pure_functions.php`
- `tests/Unit/PureFunctionsTest.php`

## Validation

This adjustment was validated with `encryptClientServer` both enabled and disabled, including:

- saving a password containing `<`
- saving a password containing the literal string `&lt;`
- displaying the password
- opening the edit form
- saving without modifying the password
- checking password history
- confirming there is no regression on the already fixed paths

PR #5210, opened 2026-05-09
Fix API item limit query parameter type casting

guerricv:
This PR fixes a TypeError in the API item endpoints when the optional `limit`
query parameter is provided.

`limit` is read from query string parameters as a string, but
ItemModel::getItems() expects an integer for its `$limit` argument.

Changes:
- Cast `limit` to int in ItemController::inFoldersAction()
- Apply the same cast in ItemController::getAction() for label searches

Test:
- php -l api/Controller/Api/ItemController.php


nilsteampassnet:
Thank you.
Merged

PR #5211, opened 2026-05-10
Fix Personal Folder Descendant Access in API

guerricv:
## Summary

This PR completes the API security fix around personal folder visibility and folder usage validation.

It builds on the existing API hardening from PR #5210 and adds server-side checks to prevent API users from seeing or using folders located under another user's personal folder, including descendant folders where `personal_folder` is not directly set.

## Bug

Personal folder descendants may have `personal_folder = 0` in `nested_tree`.

The API was previously relying too much on direct checks such as:

```sql
nt.personal_folder = 1
```

or treating folders with `personal_folder = 0` as non-personal.

As a result, a descendant folder located under another user's personal folder could still be considered visible or usable by the API if it appeared in the accessible folder list through roles or cached folder data.

This could affect:

- `folder/writableFolders`
- item creation
- item update
- item folder changes / moves
- moves between shared folders and personal folders
- moves between personal folders owned by different users

## Root Cause

The ownership check was based on the folder row itself, instead of determining whether the folder is located inside a personal folder root.

For descendants, the direct `personal_folder` flag is not reliable enough. The correct check must use the nested tree boundaries:

- `nleft`
- `nright`

A folder must be allowed only if:

- it is outside any personal folder root, or
- it is inside the current user's personal folder root.

It must be rejected if it is inside anoather user's personal folder root.

## Changes

This PR adds a centralized API folder access validator.

The new validation:

- normalizes folder and item ID lists before access checks;
- filters API folder lists using `nested_tree.nleft` / `nested_tree.nright`;
- allows shared and non-personal folders according to existing permissions;
- allows the current user's personal folder and descendants;
- rejects folders located under another user's personal folder root;
- applies the same server-side validation to item creation and update flows;
- validates both source and target folders during item moves;
- prevents moves into another user's personal folder;
- prevents moves from personal folders to unauthorized destinations;
- prevents moves between personal folders owned by different users;
- keeps folder lookup requests fail-closed when parameters are missing or unauthorized.

`writableFolders` no longer relies only on the direct `personal_folder` value of the folder row. It now checks whether the folder is under a personal root owned by another user.

Item creation and item moves now also treat descendants of personal folders as personal, even when the descendant row itself has `personal_folder = 0`.

## Security Impact

After this change:

- an API user cannot see descendants of another user's personal folder;
- an API user cannot create an item in another user's personal folder descendant;
- an API user cannot move an item into another user's personal folder descendant;
- an API user cannot move items between personal folder trees owned by different users;
- shared folders and regular non-personal folders keep their existing behavior;
- legitimate operations inside the current user's own personal folder tree remain allowed.

## PR #5210 Compatibility

The existing API behavior introduced by PR #5210 is preserved.

In particular:

- `folders_list` is still kept out of the JWT payload;
- the API still relies on `cache_tree` to store and reload accessible folders;
- `inFoldersAction()` still intersects requested folders with the authorized folder list;
- folder IDs used in API constraints are still normalized as strict integers.

This PR adds the missing personal-root ownership validation on top of that behavior.

## Testing Status

Testing is currently in progress on my side.

I am opening this PR early so the work can already be reviewed and so the testing session can be shared.

Initial checks performed:

- reviewed API folder visibility logic;
- reviewed item creation and update folder validation;
- reviewed item move source and target validation;
- confirmed the existing PR #5210 API behavior is preserved;
- added focused unit coverage for the new folder access helper behavior.

Further functional testing is still ongoing, especially around:

- `folder/writableFolders`;
- item creation in shared folders;
- item creation in the current user's personal folder descendants;
- item creation rejection in another user's personal folder descendants;
- item moves between shared and personal folders;
- item moves between personal folders owned by different users.

### DIFF
--- a/app/api/Controller/Api/FolderController.php
+++ b/app/api/Controller/Api/FolderController.php
@@ -156,39 +156,44 @@ class FolderController extends BaseController
 
         if (strtoupper($requestMethod) === 'GET') {
             try {
-                $userFolders = !empty($userData['folders_list']) ? explode(',', $userData['folders_list']) : [];
-                $rows = DB::query(
-                    'SELECT nt.id AS folder_id, nt.title, nt.nlevel, nt.parent_id
-                    FROM ' . prefixTable('nested_tree') . ' AS nt
-                    LEFT JOIN ' . prefixTable('nested_tree') . ' AS personal 
-                        ON personal.personal_folder = 1 
-                        AND personal.title = %s
-                    WHERE nt.id IN %li
-                    AND (
-                        nt.personal_folder = 0
-                        OR (
-                            personal.id IS NOT NULL
-                            AND nt.nleft >= personal.nleft 
-                            AND nt.nright <= personal.nright
-                        )
-                    )
-                    GROUP BY nt.id, nt.title, nt.nlevel, nt.parent_id
-                    ORDER BY nt.nlevel ASC, nt.title ASC',
-                    $userData['id'],
-                    $userFolders
+                $folderAccessModel = new FolderAccessModel();
+                $userFolders = $folderAccessModel->filterFoldersForUser(
+                    $folderAccessModel->normalizeFolderIds($userData['folders_list'] ?? []),
+                    (int) $userData['id']
                 );
 
                 $userId = (string) $userData['id'];
                 $username = $userData['username'];
                 $writableFolders = [];
-                foreach ($rows as $row) {
-                    $writableFolders[] = [
-                        'id' => (int) $row['folder_id'],
-                        'label' => $row['title'] === $userId ? $username : $row['title'],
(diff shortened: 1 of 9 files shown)


guerricv:
Additional note about browser extension compatibility:

While testing a backport of this API hardening on TeamPass 3.1.7.6, I noticed that the browser extension may currently interpret some API `403 Access denied` responses as `TOKEN_EXPIRED`.

This does not appear to be caused by the security fix itself on `evol/3.2.0.0`, where the PR #5210 API session-key changes are already present. However, this PR intentionally makes the API reject folders/items located under another user's personal folder, so clients that previously received such resources may now get a proper access-denied response.

The extension may therefore need a separate follow-up adjustment to distinguish a real expired token from an authorization refusal.

Issue #5236, opened 2026-06-04
[BUG] Docker is still trying to install Teampass


michaldejmek:
## Steps to reproduce

1. docker compose down && docker compose up -d

Every time Docker is stopped and restarted, the application tries to reinstall itself.

I would expect it to detect an existing instance and, at most, perform a background check.

## Server configuration

| **Deployment** | Docker sha-c9e9ca8

## Important informations

| **Updated from an older Teampass or fresh install** | No |
| **Using Docker** | Yes

michaldejmek:
@nilsteampassnet  Every time Docker stops and restarts, it deletes the app/config/ folder :( 

[version](image: teampass/teampass:sha-dba078b)

PR #5244, opened 2026-06-09
Add TeamPass Helm Chart

felipefrx:
Creation of TeamPass Helm Chart.


nilsteampassnet (changes requested):
Thanks a lot for this contribution, @felipefrx
A proper Helm chart is a great addition and the overall structure is clean (standard scaffold, recommended `app.kubernetes.io/*` labels, both Ingress and Gateway API/HTTPRoute support, HPA, PVC and a Helm test). The env-var mapping (`INSTALL_MODE`, `ADMIN_EMAIL`, `ADMIN_PWD`, `TEAMPASS_URL`, `DB_*`, `PHP_*`) matches exactly what `docker/docker-entrypoint.sh` expects, which is good.

Before we can merge, there are a few blocking points; mostly because TeamPass is a password manager, so a misconfigured deployment can mean **unrecoverable data loss** or a wide-open admin account.

## Blocking

### 1. Persistence paths do not match the official image → loss of the encryption master key
`values.yaml` mounts the volume on `/var/www/html/public/{sk,files,upload,config,secrets}`, but the official entrypoint (`docker/docker-entrypoint.sh`) stores persistent data in:
- `/var/www/html/storage/{sk,files,upload,config,backups}`
- `/var/www/html/secrets` ← the Defuse master key / `SECUREFILE`

With the official image, none of those paths are on the PVC. In particular `/var/www/html/secrets` would **not** be persisted, so the encryption key is regenerated on every pod restart and **all stored passwords become unrecoverable** (this is the same class of problem as #5236). Please align the `volumeMounts`/`subPath`s with the `storage/` + `secrets/` layout above.

### 2. Use the official, versioned image
`image.repository: felipefrx/images` with `tag: "teampass"` points to a personal Docker Hub repo and a floating tag. Please switch to the official image (`teampass/teampass`) and pin a real version tag so deployments are reproducible and maintainable. (This is also why the paths in #1 currently "work", the chart was built against a custom image whose internal layout differs from the official one.)

### 3. Admin password is exposed in a ConfigMap
`ADMIN_PWD` is under `config.variables`, so it is rendered by `configmap.yaml` in clear text (not even base64). Combined with `INSTALL_MODE: auto` and the defaults `ADMIN_EMAIL: teampass@…` / `ADMIN_PWD: teampass`, every default install creates an admin account `teampass` / `teampass`. Please move `ADMIN_PWD` into the `Secret`, and avoid shipping trivial default credentials (generate random secrets, e.g. `randAlphaNum` + `lookup` for idempotency, or make them `required`).

### 4. Weak and duplicated DB credentials
`DB_PASSWORD`, `MARIADB_ROOT_PASSWORD`, `mariadb.auth.rootPassword` and `mariadb.auth.password` are all `"teampass"`, and the DB credentials are defined twice (the `secrets` block and the `mariadb.auth` block) with no link between them - changing one side desynchronizes the app and the database. Please derive them from a single source / reuse the Bitnami subchart secret, and drop the trivial defaults.

## Non-blocking, but worth addressing

5. **HPA vs storage/app model** - `autoscaling.maxReplicas: 10` with a `ReadWriteOnce` PVC cannot work across nodes (multi-attach), and TeamPass is not designed for horizontal scaling (filesystem-based secrets/sessions). Please remove the HPA or clearly document it as unsupported.
6. **Hardcoded `DB_HOST` / URL** - `DB_HOST: teampass-mariadb.teampass` hardcodes both the release name and the namespace; the chart only works if installed as release `teampass` in namespace `teampass`. Please template it (e.g. `{{ .Release.Name }}-mariadb.{{ .Release.Namespace }}`). Same for `TEAMPASS_URL: http://localhost:8080`, which is not reachable in-cluster.
7. **Vendored dependency** - `charts/mariadb-25.1.3.tgz` is committed. Conventionally the `.tgz` is resolved via `helm dependency update` (kept out of VCS) and locked through `Chart.lock`. Also note the Bitnami `charts.bitnami.com/bitnami` catalog is being deprecated/migrated, which may affect long-term availability.
8. **Hardening** - `podSecurityContext` / `securityContext` are empty (no `runAsNonRoot`, `readOnlyRootFilesystem`, `drop: [ALL]`), and `serviceAccount.automount: true` even though TeamPass does not call the K8s API (set it to `false`). For a security product, hardened defaults are expected.
9. **Probes** - `livenessProbe` / `readinessProbe` are empty; please add at least an HTTP readiness probe on `/`.
10. **Scaffold leftovers** - `Chart.yaml` description is the placeholder "A Helm chart for Kubernetes"; `appVersion: "0.1.0"` instead of the real app version; the HTTPRoute example rule uses `value: /headers` instead of `/`; `test-connection.yaml` uses `image: busybox` without a tag; and there is no chart `README.md`.
11. **Naming consistency** - `configmap.yaml` / `secret.yaml` / `pvc.yaml` use `{{ .Release.Name }}-config|secret|data` while the other resources use the `teampass.fullname` helper; worth unifying on the helper.

Happy to help review the next revision. Thanks again for working on this!

PR #5246, opened 2026-06-10
Fix api session visibility and revocation in connected users administration

guerricv:
## Summary

This PR improves how TeamPass detects, displays, and revokes active API sessions in the administration connected-users views.

The `utilities.database` tab "Currently connected users" now detects API users from the server-side API session state instead of rebuilding presence from `log_system` entries. The same API-session awareness is applied to the database restore strict-mode modal and to the CLI restore safety checks.

The change addresses two related issues:

- API users could disappear from the connected-users tab too early because `api_token_duration` was interpreted as seconds instead of minutes;
- connected-user detection was expensive because it scanned and aggregated `log_system`, including a non-index-friendly `LIKE '%tp_src=api%'` condition.

It also tightens API session revocation: disconnecting a user from the administration interface now invalidates the server-side API session state used by JWT validation.

## Root cause

The API JWT duration setting is configured and documented in minutes.

The JWT issuer uses:

```php
min(max((int) ($SETTINGS['api_token_duration'] ?? 60), 1), 1440) * 60
```

This means a value of `60` represents a token valid for 60 minutes.

The connected-users datatable and restore script used the same setting as if it was already expressed in seconds:

```php
time() - ($apiTokenDuration + 600)
```

With the default value of `60`, the administration page only considered the API user connected for about 11 minutes instead of 70 minutes including the existing 10-minute grace period.

The previous API presence detection also depended on `log_system` rows created when a JWT was issued. This had two drawbacks:

- active API users were detected from historical logs instead of the current server-side API session state;
- every connected-users refresh could query and aggregate a potentially large log table.

In addition, the administration disconnect actions cleared the web session fields in `teampass_users`, but did not clear the API session data stored in `teampass_api`. A JWT could therefore remain valid until expiration, especially for endpoints that only checked the JWT signature and expiration.

### Connected-users datatable

The `users_logged_in` datatable now detects API sessions from `teampass_api`:

- `session_key IS NOT NULL`
- `session_key != ''`
- `timestamp >= current_time - api_token_duration_minutes - 10 minutes grace`

The API duration is now normalized consistently:

```php
$apiTokenDurationMinutes = min(max((int) ($SETTINGS['api_token_duration'] ?? 60), 1), 1440);
$apiConnectedAfter = time() - (($apiTokenDurationMinutes * 60) + 600);
```

The datatable still displays both connection types:

- active web session from `users.session_end`;
- active API session from `api.timestamp` and `api.session_key`.

When a web session has expired but the API session is still active, the "connected since" duration uses the API session timestamp.

### Restore strict-mode modal

The restore connected-users check now counts both web and API sessions.

This keeps the restore modal aligned with the connected-users tab and prevents the operator from continuing a restore while an API user is still considered active.

### CLI restore safety check

The CLI restore script now uses the same API-session source as the administration UI.

Without `--force-disconnect`, the script blocks when either web or API sessions are active and records:

```text
active_sessions
```

With `--force-disconnect`, the script now clears active API sessions in addition to web sessions.

The restore initiator is also disconnected from both web and API session state before the restore starts.

### User disconnect actions

The existing connected-users disconnect actions now invalidate API session state as well as web session state.

The new helper clears transient API session fields:

- `encrypted_private_key`
- `session_key_salt`
- `session_key`
- `session_aes_key`
- `timestamp`

The permanent API configuration remains untouched, including API permissions and API key configuration.

### JWT validation

JWT validation now checks the server-side API session state after verifying the JWT signature and expiration.

A JWT is only considered valid when:

- it contains a valid user id;
- it contains a non-empty `key_tempo`;
- `teampass_api.session_key` matches the JWT `key_tempo`;
- `session_aes_key` is present;
- `encrypted_private_key` is present.

This makes administration-side API session invalidation effective for all API endpoints that pass through the shared `verifyAuth()` path, not only endpoints that later decrypt the user's private key.

### API authentication fallback

When issuing a JWT, the API now updates the user's `teampass_api` row as before.

If no row exists for the user, a fallback insert creates the missing API session row with the user's current API permissions. This avoids issuing a JWT that immediately fails the new server-side session validation on older or inconsistent installations.

## User-facing behavior

After this change, administrators should see API users in "Currently connected users" for the same duration as the issued JWT remains valid, plus the existing 10-minute grace period.

For example, with `api_token_duration = 60`, an API user is considered connected for about 70 minutes after JWT issuance, instead of about 11 minutes.

The API indicator in the connected-users table continues to show which users are currently considered connected through API activity.

Disconnecting a user from the connected-users tab now also invalidates their API session. The user must authenticate again to obtain a valid API JWT.

## Restore behavior

The restore strict-mode modal and CLI restore now have consistent connected-user semantics.

When strict mode checks connected users:

- web sessions are counted from `users.session_end`;
- API sessions are counted from `teampass_api.session_key` and `teampass_api.timestamp`;
- the current administrator is still excluded where applicable.

When the operator chooses or runs force disconnect:

- web session fields are cleared;
- API session fields are cleared;
- active JWTs tied to those API sessions stop passing global JWT validation.

## Performance impact

The connected-users tab no longer needs to scan and aggregate `log_system` to detect API presence.

The previous logic used:

- `EXISTS` against `log_system`;
- a grouped `MAX(date)` subquery over `log_system`;
- a `field_1 LIKE '%tp_src=api%'` condition.

The new logic uses `teampass_api`, which already has:

- a `user_id` key;
- an `idx_api_timestamp` index.

This should reduce the cost of loading the connected-users tab on installations with large log tables.

## Security impact

The change improves API session revocation.

Before this PR, an administrator disconnect mostly affected the web session fields. API JWTs could remain accepted until expiration if the endpoint only validated the JWT signature and `exp` claim.

After this PR, API JWT validity also depends on server-side session material in `teampass_api`. Clearing that session state immediately invalidates the JWT.

The private key protection model remains unchanged: the API still stores the encrypted private key and session AES key server-side, and the JWT still only carries the `key_tempo` reference.

## Files changed

- `app/sources/logs.datatables.php`
  - Connected-users API detection now uses `teampass_api`.
  - `api_token_duration` is interpreted as minutes.

- `app/sources/backups.queries.php`
  - Restore connected-users modal count now includes active API sessions.

- `app/sources/users.queries.php`
  - User disconnect actions now clear API session state.
  - Disconnect-all now includes active API users, not only web users.

- `app/scripts/restore.php`
  - Restore CLI detects active API sessions from `teampass_api`.
  - Restore CLI blocks on API sessions unless `--force-disconnect` is used.
  - Force disconnect now clears API session state.

- `app/api/inc/jwt_utils.php`
  - JWT validation now checks the server-side API session state.

- `app/api/Model/AuthModel.php`
  - JWT issuance updates server-side API session state.
  - Missing API rows are inserted as a fallback for older or inconsistent user records.

## Testing

Recommended validation:

- set `api_token_duration` to 60 minutes;
- authenticate through the API or browser extension;
- open `utilities.database` and the "Currently connected users" tab;
- confirm the API user remains visible beyond the previous short window;
- confirm the API indicator is displayed for the user;
- disconnect the API user from the connected-users tab;
- confirm the row disappears after refresh;
- retry an API call with the old JWT and confirm it is rejected;
- re-authenticate through the API and confirm a new JWT works;
- open the restore strict-mode modal while an API session is active;
- confirm the modal blocks continuation;
- use "disconnect all" from the modal and confirm API sessions are cleared;
- run the restore CLI without `--force-disconnect` while API users are active and confirm it exits with active-session errors;
- run the restore CLI with `--force-disconnect` and confirm it clears web and API sessions before continuing.

Technical checks attempted locally:

- PHP syntax linting was attempted with `php -l`, but PHP CLI was not available in the local Windows PATH or common local PHP installation paths.

## Notes

This PR does not implement per-request API heartbeat tracking.

API presence is still based on the last successful JWT issuance and its configured validity window. This matches the current API authentication model and avoids adding a database write on every API request.

A future enhancement could add throttled `last_api_seen_at` updates during authenticated API requests if real-time API activity, rather than valid JWT session presence, becomes necessary.

### DIFF
--- a/app/api/Model/AuthModel.php
+++ b/app/api/Model/AuthModel.php
@@ -260,18 +260,41 @@ class AuthModel
         // Store the ENCRYPTED private key and the AES session key server-side.
         // The session_aes_key never leaves the server (not embedded in JWT),
         // so a stolen JWT alone cannot decrypt the private key.
-        DB::update(
+        $apiSessionData = [
+            'encrypted_private_key' => $encryptedPrivateKey,
+            'session_key_salt' => $sessionKeySalt,
+            'session_key' => $keyTempo,
+            'session_aes_key' => base64_encode($sessionKey),
+            'timestamp' => time(),
+        ];
+
+        $apiSessionUpdated = DB::update(
             prefixTable('api'),
-            [
-                'encrypted_private_key' => $encryptedPrivateKey,
-                'session_key_salt' => $sessionKeySalt,
-                'session_key' => $keyTempo,
-                'session_aes_key' => base64_encode($sessionKey),
-                'timestamp' => time(),
-            ],
+            $apiSessionData,
             'user_id = %i',
             $userInfo['id']
         );
+
+        if ((int) $apiSessionUpdated === 0) {
+            DB::insert(
+                prefixTable('api'),
+                array_merge(
+                    [
+                        'type' => 'user',
+                        'label' => null,
+                        'value' => null,
+                        'user_id' => (int) $userInfo['id'],
+                        'allowed_folders' => (string) ($userInfo['api_allowed_folders'] ?? ''),
+                        'enabled' => (int) ($userInfo['api_enabled'] ?? 1),
+                        'allowed_to_create' => (int) ($userInfo['api_allowed_to_create'] ?? 0),
+                        'allowed_to_read' => (int) ($userInfo['api_allowed_to_read'] ?? 1),
+                        'allowed_to_update' => (int) ($userInfo['api_allowed_to_update'] ?? 0),
+                        'allowed_to_delete' => (int) ($userInfo['api_allowed_to_delete'] ?? 0),
+                    ],
+                    $apiSessionData
+                )
+            );
+        }
 
         // get user folders list and persist in cache_tree
         $ret = $this->buildUserFoldersList($userInfo);
--- a/app/api/inc/jwt_utils.php
+++ b/app/api/inc/jwt_utils.php
@@ -87,11 +87,11 @@ function is_jwt_valid($jwt) {
 		$decoded = (array) JWT::decode($jwt, new Key(getApiJwtSigningKey(), 'HS256'));
 
 		// Check if expiration is reached
-		if ($decoded['exp'] - time() < 0) {
+		if (!isset($decoded['exp']) || $decoded['exp'] - time() < 0) {
 			return false;
 		}
 
-		return true;
+        return is_api_session_valid($decoded);
 	} catch (InvalidArgumentException $e) {
 		// provided key/key-array is empty or malformed.
 		return false;
@@ -118,6 +118,33 @@ function is_jwt_valid($jwt) {
 		// provided key ID in key/key-array is empty or invalid.
 		return false;
 	}
+}
+
+function is_api_session_valid(array $decoded): bool
+{
+    $userId = (int) ($decoded['id'] ?? 0);
+    $keyTempo = (string) ($decoded['key_tempo'] ?? '');
+
+    if ($userId <= 0 || $keyTempo === '') {
+        return false;
+    }
+
+    $apiSession = DB::queryFirstRow(
+        'SELECT session_key, session_aes_key, encrypted_private_key
+         FROM ' . prefixTable('api') . '
+         WHERE user_id = %i',
+        $userId
+    );
+
+    if ($apiSession === null || empty($apiSession['session_key'])) {
+        return false;
+    }
+
+    if (!hash_equals((string) $apiSession['session_key'], $keyTempo)) {
+        return false;
+    }
+
+    return !empty($apiSession['session_aes_key']) && !empty($apiSession['encrypted_private_key']);
 }
 
 function base64url_encode($data) {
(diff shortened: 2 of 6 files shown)


nilsteampassnet:
### Adaptation note (target branch `develop`)

`develop` already ships a per-token API session model (`teampass_api_sessions`, one row per JWT keyed by `jti`, with per-request revocation in `api/index.php` and a `/auth/logout` endpoint). This PR was written against `master`, which only has the single-row `teampass_api` model.

To port it to `develop`, the API-core changes were re-based on `api_sessions`:

- **Kept (user-facing goal):** API users are now detected for the full JWT validity window instead of ~11 min, the `log_system` scan is dropped, and the admin disconnect / restore (UI + CLI) now revoke API sessions.
- **Detection/revocation source:** `teampass_api_sessions` (`revoked_at IS NULL AND expires_at >= now`) instead of the single-row `teampass_api`. This is required because `develop` supports concurrent API clients per account and per-token profile revocation.
- **Dropped:** the `is_jwt_valid()` / `AuthModel` fallback-insert changes; on `develop` they are redundant (revocation is already enforced for every endpoint in `api/index.php`) and would break concurrent sessions.
- **Behavioral delta:** the "+10 minutes grace" is removed. The connected window now equals the JWT's real validity, because expired JWTs are hard-rejected on `develop` (so a "connected" row past expiry would be misleading). With `api_token_duration = 60`, an API user shows as connected for ~60 minutes.

Issue #5212, opened 2026-05-12
[BUG] - Renewal page not working


Berzekerz:
## Page on which it happened
index.php?page=options

I’m trying to use the renewal function on the folders, but when I check the renewal page, nothing appears.

No matter what value I set, no folders are displayed on the renewal page. I tested with 1, 5, 10, and 30 days, and nothing shows up. I also tried applying the number of days using both the admin account and a regular user account, but the behavior is the same.

I noticed the option says to configure the number of days before activating the feature, but since this installation has been running for a while, i’m not sure if it may have been configured it incorrectly in the past, and because of that it’s not working now.

<img width="1655" height="897" alt="Image" src="https://github.com/user-attachments/assets/381b6d49-a884-4502-b639-5f05bede94d1" />

<img width="397" height="785" alt="Image" src="https://github.com/user-attachments/assets/4af5739b-238a-4e5c-8d07-adf32b9d4ce5" />

<img width="1035" height="506" alt="Image" src="https://github.com/user-attachments/assets/183879b1-9add-409e-bc73-ae91b855b172" />

## Server configuration

| | |
|---|---|
| **Operating system** | Linux quality-teampass-01 6.12.63+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.63-1 (2025-12-30) x86_64 |
| **Web server** | Apache/2.4.66 (Debian) |
| **Database** | 11.8.3-MariaDB-0+deb13u1 from Debian |
| **PHP version** | 8.4.16 |
| **TeamPass version** | 3.1.7.6 |
| **Install type** | Fresh install / Updated from x.x.x |
| **Deployment** | Standard / Docker / other |

## TeamPass configuration variables

Configuration variables (click to expand)

```
'activate_expiration' => '1'
'admin_2fa_required' => '1'
'agses_authentication_enabled' => '0'
'allow_import' => '0'
'allow_print' => '1'
'anyone_can_modify' => '0'
'anyone_can_modify_bydefault' => '0'
'api' => '0'
'api_token_duration' => '60'
'backup-scheduled-enabled' => '1'
'bck_script_filename' => 'bck_teampass'
'bck_script_passkey' => ''
[...]

PR #5247, opened 2026-06-11
Fix 5212 Renewal page expiration listing

guerricv:
## Summary

This PR fixes the Renewal utility page so it lists items whose passwords are expired or will expire by the selected date.

It addresses the behavior reported in issue #5212, where the Renewal page stayed empty even when folders had a renewal delay configured.

It also cleans up the Renewal page table by removing the item-open link, which is not useful for administrators who cannot open item cards from this utility view.

## Root cause

The Renewal datatable was reading from `automatic_del` with `del_type = 2`.

That table belongs to the "delete after consultation" feature. It is populated when an item is configured for automatic deletion after a number of views or after a deletion date.

Password renewal, however, is configured on folders through `nested_tree.renewal_period`. The expiration date must be computed from:

- the item's creation date;
- or the most recent password-change log;
- plus the folder renewal period.

The page also had a secondary access check mismatch: `utilities.renewal.js.php` checked `utilities.logs` instead of `utilities.renewal`.

## Implementation details

The Renewal datatable now queries active items from visible folders and computes expiration directly from TeamPass renewal data:

- `items` for active item metadata;
- `nested_tree.renewal_period` for the folder delay;
- `log_items` for the latest relevant date, using `at_creation` or `at_modification` with `at_pw%`.

The selected date is treated as an inclusive target day. If no date is selected, the page lists items already expired at the current time.

The calculation now consistently treats `renewal_period` as days by using `TP_ONE_DAY_SECONDS`, matching the folder UI and documentation.

The Renewal datatable now preserves DataTables request parameters when adding the selected date filter, so server-side search, ordering, and pagination keep working.

Related expiration indicators were aligned in:

- the item list;
- search results;
- cache refresh/add/update helpers used by search.

The Renewal utility page layout was refreshed with the existing AdminLTE card, callout, input group, and responsive table patterns used elsewhere in TeamPass.

## User-facing behavior

After this change:

- folders with a renewal delay greater than `0` produce Renewal page results;
- changing the date picker shows items expiring up to that date;
- using the Renewal table search field filters the server-side results;
- Renewal rows show the item label, expiration date, and folder path without an unusable item-open link;
- the item list and search expiration flags use the same day-based renewal delay;
- users allowed to access `utilities.renewal` no longer need `utilities.logs` just to load the Renewal page JavaScript.

## Testing

Manual/static validation covered:

- verifying that `expired.datatables.php` no longer depends on `automatic_del`;
- verifying that renewal expiration uses `nested_tree.renewal_period` and `log_items`;
- verifying that legacy/current password-change log reasons are matched through `at_pw%`;
- verifying that no renewal calculation still uses `TP_ONE_MONTH_SECONDS`.
- verifying that `expired.datatables.php` no longer emits the item-open link column.

## Notes

This PR does not change the automatic deletion feature. Existing `automatic_del` usage remains limited to delete-after-consultation workflows.

### DIFF
--- a/app/pages/utilities.renewal.js.php
+++ b/app/pages/utilities.renewal.js.php
@@ -71,7 +71,7 @@ $checkUserAccess = new PerformChecks(
 );
 // Handle the case
 echo $checkUserAccess->caseHandler();
-if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPage('utilities.logs') === false) {
+if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPage('utilities.renewal') === false) {
     // Not allowed page
     $session->set('system-error_code', ERR_NOT_ALLOWED);
     include TEAMPASS_ROOT . '/public/error.php';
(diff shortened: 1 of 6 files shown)


nilsteampassnet:
Merged in develop

PR #5281, opened 2026-07-12
tests: Add basic security check for API folder endpoint

ZaidMansuri07:
Hi,
I have added a basic integration test case inside the `tests/` directory to verify that the API properly drops or handles incoming requests trying to create folders when a valid API token is missing. 

This helps expand the test framework layout as the project stabilizes code for newer PHP environments.

### DIFF
--- a/tests/FolderCreationTest.php
+++ b/tests/FolderCreationTest.php
@@ -1,0 +1,17 @@
+'add_folder',
+            'title' => 'Test DevOps Folder'
+        ]);
+
+        $ch = curl_init($this->apiUrl);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_POST, true);
+        curl_setopt($ch, CURLOPT_POSTFIELDS, $payload);
+        curl_setopt($ch, CURLOPT_HTTPHEADER, ['Content-Type: application/json']);
+
+        $response = curl_exec($ch);
+        $httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+        curl_close($ch);
+
+        $this->assertNotEquals(200, $httpCode, "The API allowed access without a valid API key token.");
+    }
+}


nilsteampassnet:
Thanks for the contribution @ZaidMansuri07 ! Before we can consider merging, a few blocking points need clarification:

1. The file isn't valid PHP. tests/FolderCreationTest.php starts mid-statement ('add_folder',) with no <?php tag, no class, no test method signature, and $this->apiUrl / $payload are never defined. PHPUnit can't load or run it as-is. It looks like the top of the file got lost during the commit.
2. Wrong endpoint. add_folder is a session-authenticated web AJAX action (app/sources/folders.queries.php), not a token-protected REST endpoint. The API folder creation is POST /api/folder/create (FolderController.php), which takes title / parent_id / complexity, not a type field.
3. Assertion too loose. assertNotEquals(200, $httpCode) also passes on 500, 404, or 0 (server unreachable / curl failure), so it can green without actually testing anything. A missing-token request should assert the specific expected code (401).
4. Live integration test in a unit suite. The existing tests/Unit/** suite is pure unit tests (no network/DB); the API security tests use static source scanning specifically to avoid needing a running server. This test needs a live apiUrl with no CI infra to provide one.

Could you clarify what you intended to test exactly? If the goal is "unauthenticated request to the API folder endpoint is rejected", we're happy to help reshape it against POST /api/folder/create with a precise 401 assertion, but the file needs to be complete, valid PHP first.

ZaidMansuri07:
Hi @nilsteampassnet,

Thanks so much for taking the time to give such a detailed breakdown. That makes a lot of sense and I see what you mean now.

First my bad on the broken file structure. the top half of the code definitely got cut off during my commit, so i will make sure the final version is complete valid PHP.

My main goal was exactly what you guessed
making sure that if someone tries to create a folder without being logged in, the system completely blocks them with a 401 error.

I totally get your point about the unit tests and the CI environment not having a live server or database to run network requests against. since you mentioned using static scanning or pure unit tests for API security instead.
what would be the best way for me to fix this?

I can either
1. Write a proper unit test that mocks the request/controller if the testing setup supports it.
2. Help add a new check/rule to your existing static analysis setup instead.

Let me know which way fits your project best, and i will be happy to update the code

PR #5284, opened 2026-07-15
Update italian.php

danfossi:
fixed typo in login screen


danfossi:
All translations were done through POEditor, so there is no point in keeping this PR open

PR #5287, opened 2026-07-16
Feature: Add native KeePass 2.x XML export support

danfossi:
This PR adds a native export functionality to the KeePass 2.x XML format:

xml_export


danfossi:
Note:
For XML export to work properly, you need to have the php-xml package installed.

PR #5295, opened 2026-07-21
Add native KeePass 2.x XML export support

danfossi:
his PR adds a native export functionality to the KeePass 2.x XML format:

622591094-b83440c7-b905-4365-84a6-7d2c09fa5d2c

### DIFF
--- a/app/pages/export.js.php
+++ b/app/pages/export.js.php
@@ -157,7 +157,7 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
     // On type selection
     $('#export-format').on("change", function(e) {
         $('#download-export-file').attr('onclick', "").addClass('hidden');
-        if ($(this).val() === 'pdf' || $(this).val() === 'html') {
+        if ($(this).val() === 'pdf' || $(this).val() === 'html') {
             $('#pwd').removeClass('hidden');
             $('#export-password').val('');
         } else {
@@ -295,11 +295,11 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
         } else if ($('#export-format').val() === 'html') {
             // Offline mode: build a single self-contained, password-encrypted HTML file
             generateOfflineFile(ids);
-        } else if ($('#export-format').val() === 'csv') {
+        } else if ($('#export-format').val() === 'csv' || $('#export-format').val() === 'xml') {
             // Export to CSV
             $.post(
                 "sources/export.queries.php", {
-                    type: 'export_to_csv_format',
+                    type: 'export_to_' + $('#export-format').val() + '_format',
                     ids: (JSON.stringify(ids))
                 },
                 function(data) {
@@ -321,7 +321,17 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
                     data = decodeQueryReturn(data, '<?php echo $session->get('key'); ?>');
                     
                     // download VSC file
-                    download(new Blob([data.csv_content]), $('#export-filename').val() + ".csv", "text/csv");//decodeURI(data[0].content)
+                    if (data.xml_content) {
+                        var byteCharacters = atob(data.xml_content);
+                        var byteNumbers = new Array(byteCharacters.length);
+                        for (var i = 0; i < byteCharacters.length; i++) {
+                            byteNumbers[i] = byteCharacters.charCodeAt(i);
+                        }
+                        var byteArray = new Uint8Array(byteNumbers);
+                        download(new Blob([byteArray]), $('#export-filename').val() + ".xml", "text/xml");
+                    } else if (data.csv_content) {
+                        download(new Blob([data.csv_content]), $('#export-filename').val() + ".csv", "text/csv");
+                    }//decodeURI(data[0].content)
                 }
             );
         }
--- a/app/pages/export.php
+++ b/app/pages/export.php
@@ -125,6 +125,7 @@ header('Cache-Control: no-cache, no-store, must-revalidate');
                         <label><?php echo $lang->get('export_format_type'); ?></label>
                         <select class="form-control select2" style="width:100%;" id="export-format">
                             <option value="csv"><?php echo $lang->get('csv'); ?></option>
+                                   <option value="xml">KeePass XML</option>
                             <?php
                             if (isset($SETTINGS['settings_offline_mode']) === true && (int) $SETTINGS['settings_offline_mode'] === 1) {
                                 echo '<option value="html">'.strtoupper($lang->get('html')).'</option>';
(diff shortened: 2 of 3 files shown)


nilsteampassnet:
Hi @danfossi, thanks a lot for this contribution, a native KeePass 2.x XML export is a frequently requested feature, and the overall approach (DOM building, group hierarchy reconstruction, base64 transport to preserve UTF-8) is sound.

Before I can merge it, could you please rebase the work and address a few points?

## 1. Please base the PR on `develop`

This PR currently targets `master` from your fork's `master`. `master` is the stable branch, **`develop` is where evolutions land**. Please rebase your branch onto `develop` and retarget the PR there (ideally from a branch named `pr-5295`).

This matters beyond convention: your fork is based on an older `master`, and the rebase artifact is causing a real regression, see point 2.

## 2. Restore `pw_iv` in the CSV and PDF export paths (blocking)

The diff removes `i.pw_iv AS pw_iv` from two `SELECT` statements and the 4th argument `(string) ($dataItem['pw_iv'] ?? '')` from two `teampassDecryptPasswordValue()` calls in `app/sources/export.queries.php`.

`pw_iv` carries the AES v2 metadata (`version|nonce|salt`), and `doDataDecryption()` branches on it:

```php
if ($meta === '') { CryptoManager::aesDecrypt(...)     // legacy CBC
} else {           CryptoManager::aesGcmDecrypt(...) } // AES v2
```

Without it, every AES v2 encrypted item falls into the legacy CBC branch, the exception is swallowed, and **the password is exported empty**. AES v2 is enabled by default on new installations, so this would silently break CSV and PDF exports for most users. Rebasing on `develop` should resolve these 6 lines.

## 3. Use `teampassDecryptPasswordValue()` in the XML path (blocking)

```php
$pw = base64_decode(doDataDecryption($dataItem['pw'], decryptUserObjectKey(...)));
```

Same AES v2 problem, plus this bypasses `teampassNormalizeLegacyPassword()`, so legacy passwords will be incorrect. Please select `pw_len` and `pw_iv` and mirror what the CSV path does:

```php
$pw = teampassDecryptPasswordValue(
    $dataItem['pw'],
    decryptUserObjectKeyWithMigration(
        $dataItem['share_key'],
        $session->get('user-private_key'),
        $session->get('user-public_key'),
        (int) $dataItem['increment_id'],
        'sharekeys_items'
    ),
    (int) ($dataItem['pw_len'] ?? 0),
    (string) ($dataItem['pw_iv'] ?? '')
);
```

Note `decryptUserObjectKeyWithMigration()` rather than `decryptUserObjectKey()`,, it transparently upgrades phpseclib v1 sharekeys to v3 and is required in new code. The `SELECT` must then include `s.increment_id`.

## 4. Enforce the item-level `restricted_to` check (blocking, authorization)

The XML path exports every item in the folder:

```php
$rows = DB::query('SELECT ... WHERE i.inactif = 0 AND i.id_tree = %i', intval($id));
```

The CSV path applies a second per-item filter that is missing here:

```php
if ((in_array($record['id_tree'], $session->get('user-personal_visible_folders')) === true)
    || (in_array($record['id_tree'], $session->get('user-accessible_folders')) === true
        && (empty($record['restricted_to']) === true
            || in_array($session->get('user-id'), explode(';', $record['restricted_to'])) === true)))
```

Without it, a user with folder access can export in cleartext items that are restricted to other users, items they cannot even open in the UI. Please select `i.restricted_to` and apply the same check.

## 5. Log the export (blocking, audit)

CSV and PDF both call `logItems(..., 'at_export', ...)` for each exported item. The XML path logs nothing, which would make a full vault export invisible in Admin → Logs (`at_export` is an exposed filter there). Please add the same call with `'xml'` as the last argument.

## 6. Don't return the exception message to the client

```php
'message' => 'Errore fatale XML: ' . $e->getMessage()
```

MeekroDB exception messages contain the full SQL query and server paths. Please return a generic translated message and keep the detail server-side only.

## 7. Comments and strings must be in English

A few Italian comments and strings remain: `// Helper functions per lo standard KeePass`, `// Gruppo Principale (Database)`, `'Errore fatale XML: '`. The project requires English throughout.

---

## Smaller points (please fix, but non-blocking)

- **PHPStan level 4** - the PR introduces one error (`develop` is currently clean):
  `Variable $pw on left side of ?? always exists and is not nullable` - `$pw` is initialised to `''` above, so the `?? ''` is dead. You can verify with `php app/vendor/bin/phpstan analyse -c phpstan.neon app/sources/export.queries.php`.
- **`@ini_set('memory_limit', '-1')`** - unlimited memory on a user-triggerable AJAX handler is a DoS vector; no other case in the file does this. `@set_time_limit(0)` is also redundant, it's already set at the top of the file.
- **`while (ob_get_level() > 0) { ob_end_clean(); }`** - this blindly destroys buffers opened upstream (`$checkUserAccess->caseHandler()` has already echoed by then). Please remove it.
- **`error_log()` calls** - debug traces shouldn't ship; if you want them, gate them on `LOG_TO_SERVER` the way `doDataDecryption()` does.
- **JS error handling** - `toastr.success('Done')` fires unconditionally before the `data.xml_content` test, so on the error path the user sees a success toast and gets no file. Please add an early `if (data.error === true) { toastr.error(data.message); return }`.
- **Untranslated label** - `<option value="xml">KeePass XML</option>` is hardcoded while every other option uses `$lang->get()`. A new `export_format_keepass_xml` key in `app/includes/language/*.php` would be ideal. The indentation is also off by 11 spaces.
- **`htmlspecialchars(..., ENT_XML1, 'UTF-8')`** - without `ENT_SUBSTITUTE`, an invalid UTF-8 sequence makes `htmlspecialchars` return an empty string, silently dropping the password. Please use `ENT_XML1 | ENT_SUBSTITUTE`.
- **Perf** - `Uint8Array.from(byteCharacters, c => c.charCodeAt(0))` replaces the per-character loop and is much faster on large exports.

---

Nice catch, by the way: your change to `export.js.php:160` replaces a stray non-breaking space (U+00A0) that was sitting in the source. 👍

The XML export doesn't currently carry email, tags, KB entries, custom fields or TOTP - that's fine for a first iteration, no need to add them in this PR.

Thanks again, and let me know if you'd like a hand on any of these.

danfossi:
Hi Nils,
Thank you for the detailed review!

I have tried addressed all the points raised in your feedback:
1. **Rebase & Retarget**: Rebased onto `develop` and submitted a fresh PR from `pr-5295`: ##5300
2. **AES v2 metadata**: Restored `pw_iv` in the CSV, PDF, and HTML export queries to ensure AES v2 encrypted passwords decrypt correctly.
3. **XML Decryption & Migration**: Updated XML export to use `teampassDecryptPasswordValue()` and `decryptUserObjectKeyWithMigration()`.
4. **Authorization**: Enforced per-item `restricted_to` and folder visibility checks in the XML export path.
5. **Audit Logging**: Added `logItems(..., 'at_export', ..., 'xml')` logging for every exported item in XML format.
6. **Error Security**: Replaced raw SQL/MeekroDB exception leaks with generic translated user messages, logging detailed stack traces server-side.
7. **Localization & Quality**: Translated all comments to English (sorry, I use Italian to have easily findable references in the code and I forgot to translate them), added `export_format_keepass_xml` language key, fixed PHPStan level 4 issues, improved JS error handling & performance, and used `ENT_XML1 | ENT_SUBSTITUTE`.

Closing this PR in favor of ##5300

PR #5302, opened 2026-07-25
Fix GUID mapping + nested AD groups

RealPurePurple:
### Configuration note

For nested Active Directory group membership to be evaluated when using the "Restrict login to LDAP group (DN)" option, the login group check mode must use the group-centric mode:

ldap_allowed_login_group_mode = group

The user-centric mode relies on the user's memberOf attribute and usually only contains direct group memberships in Active Directory. Therefore it may not detect indirect membership through nested AD groups.

The group-centric mode uses isUserInAllowedGroup(), which this PR extends with the AD recursive matching rule:

1.2.840.113556.1.4.1941

```markdown
For existing installations where `ldap_allowed_login_group_mode` is set to `user`,
switching it to `group` may be required


RealPurePurple (commented):
My first pull on life<3

RealPurePurple:
### Additional note: multiple matched AD groups

After enabling nested AD group resolution, a user may match several AD groups that are mapped to different TeamPass roles.

In such cases, the current role assignment logic should ensure deterministic behavior. LDAP result ordering should not be used as an implicit priority, because group order is not guaranteed.

Possible expected behaviors could be:

1. assign all matching TeamPass roles to the user, or
2. introduce an explicit role/group mapping priority, or
3. document that mapped AD groups should be mutually exclusive.

Without deterministic conflict resolution, a user belonging to multiple mapped AD groups may not always receive the role with the expected highest privileges.

nilsteampassnet:
Thanks for this and congratulations on your first PR! 🎉

Your diagnosis is correct: Active Directory only populates `memberOf` with direct memberships, so nested groups were not resolved. That issue has already been fixed on `develop` and `master` in commits `9c8b4ec` (*Fix AD group GUID mapping and add nested group support*) and `73c392c` (*Resolve nested AD groups in the user-centric login group mode*), which are ancestors of the branch this PR now targets. As a result the diff here sits **on top of** the existing fix rather than providing it, so I'm going to close it, but a few notes so nothing is lost:

**On the configuration note in your description** - it is no longer needed. `73c392c` removed the limitation you describe: in `ldap_allowed_login_group_mode = user`, when `memberOf` does not carry the group, `ActiveDirectoryExtra` now reads the *user* entry with a scope=base search filtered on `(memberof:1.2.840.113556.1.4.1941:=<groupDN>)`, the reverse form of the rule you used on the group side. Both modes behave identically now, and the setting is a pure performance trade-off. No reconfiguration is required on existing installations.

**On the current `isUserInAllowedGroup()` implementation** - the nested probe lives in a dedicated `isNestedMemberOfGroup()` helper called **last**, after the `member` / `uniqueMember` / `memberUid` checks, for two deliberate reasons:

- `LDAP_MATCHING_RULE_IN_CHAIN` is an unindexed transitive search, so a direct member never pays for it;
- the helper owns its own `try/catch`, so a directory that rejects the extended matching rule (OpenLDAP returns `inappropriateMatching`) falls back to the attribute-only answer instead of denying the login. In this PR the probe runs inside the outer `try`, so that exception would propagate to the outer `catch` and return `false`, a legitimate direct member would be refused before ever reaching the fallback the comment describes. Worth keeping in mind for future work in this area.

**On your follow-up comment about multiple matched AD groups** - good instinct, but the current behaviour is already option 1 in your list. `identify.php` accumulates *every* matching role into `$user_ad_groups`, persists them all through `setUserRoles(..., 'ad')`, and builds `fonction_id` as their union. There is no dependency on LDAP result ordering and no "first match wins" path, so no non-deterministic privilege assignment.

Two smaller points: the title mentions a GUID mapping fix, but the diff contains no GUID change - that part is already in the base (`9c8b4ec` aligned both conversion sites on `LdapRecord\Models\Attributes\Guid` and shipped the `ldap_groups_roles` migration in `upgrade_run_3.2.2.php`). And the missing newline at end of file is a genuine catch, I'll fix it separately, on **both** copies of the class.

That last point is the one thing to know for any future contribution: every `teampassclasses` package exists in both `app/includes/libraries/teampassclasses/` and `app/vendor/teampassclasses/`, and **only the `vendor/` copy is autoloaded**, editing the `includes/libraries/` one alone produces a change with no runtime effect. `tests/Unit/LdapExtraCopiesInSyncTest.php` guards against exactly that; running `php app/vendor/phpunit/phpunit/phpunit tests/Unit/` before pushing catches it. Also please target `develop` rather than `master` for new work.

Thanks again for taking the time to report and fix this, the analysis was on point, it just landed in parallel with the same fix. Contributions very welcome.

PR #5312, opened 2026-07-28
Remove user-manager ability to assign all roles

joeshubaccount:
Admins (user-admin) and "Teampass managers" (user-can_manage_all_users) should see all roles.

The if() clause also included "Managers" (user-manager) but these should only see their own roles.

EDITS:
Removed the OR for user-manager so now they can only add roles that they are a member of.
This has been tested and appears to work as intended.


nilsteampassnet:
Thank you

PR #5316, opened 2026-08-03
Fix legacy HTML entities displayed in Utilities logs

guerricv:
## Summary

This change fixes database-backed labels, folder names, logins, first names, and last names being displayed as literal HTML entities on `index.php?page=utilities.logs`.

Examples of the affected output included:

- `Cl&eacute;mence` instead of `Clémence`;
- `Code acc&egrave;s` instead of `Code accès`.

The issue was visible in the **Connections** and **Items** tabs and could affect the other log tabs using the same DataTables rendering pipeline.

## Root cause

Some current or legacy TeamPass records contain named or numeric HTML entities. The log endpoints escaped those values with `htmlspecialchars()` before JSON serialization. When an already encoded value was escaped again, for example `&eacute;` becoming `&amp;eacute;`, the client-side renderer removed only one encoding layer.

The remaining entity then had to stay escaped because the DataTables renderer deliberately treats database values as text to prevent stored XSS. Consequently, the entity source was displayed literally.

This behavior became visible after the log-table XSS hardening correctly stopped returning decoded database content as active HTML.

### Shared server-side normalization

`normalizeLogDisplayValue()` was added to `app/sources/main.functions.php`.

The helper:

1. decodes one stored HTML-entity layer with `ENT_QUOTES | ENT_HTML5` and UTF-8;
2. escapes the decoded value with `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML5`;
3. disables double encoding so a remaining legitimate entity layer is preserved for the client renderer.

This handles raw UTF-8 values, single-encoded legacy values, and values that have already been encoded twice.

### Utilities log data sources

`app/sources/logs.datatables.php` now applies the shared helper to database-backed text sent to the following views:

- Connections;
- item access/copy logs;
- Administration;
- Items;
- authentication lockout user display;
- Failed connections;
- Errors;
- background task user display values returned by the same endpoint.

The raw authentication-lockout key remains unchanged in the JSON response because it is an operational identifier used by the unlock action. Only its user-facing label is normalized.

### Knowledge base logs

`app/sources/kb.queries.php` uses the same helper for the Knowledge base log label, user login, action, and reason fields.

### Client-side rendering

The existing DataTables sequence remains in place:

1. decode the remaining display entity layer;
2. insert the result through jQuery's text API;
3. return escaped HTML to DataTables.

The authentication-lockout renderer was aligned with this sequence so user display names receive the same treatment as the other tabs.

## Security considerations

The stored-XSS protection is intentionally preserved:

- database values are JSON encoded;
- decoded values are never returned directly as active HTML;
- the final client-side insertion uses `text(...)` before DataTables receives the cell content;
- raw and entity-encoded markup therefore remains inert text;
- operational database identifiers are not modified.

The fix does not revert the previous XSS hardening and does not require a database migration.

## Tests added

`tests/Unit/UtilitiesLogsEncodingTest.php` adds regression coverage for:

- raw UTF-8 accents;
- named HTML entities;
- numeric HTML entities;
- double-encoded entities;
- ordinary ampersands;
- raw, encoded, and double-encoded HTML payloads;
- use of the normalizer by all Utilities log data sources;
- the client-side decode-then-escape contract;
- preservation of the raw authentication-lockout identifier.

## Suggested manual validation

1. Open **Utilities > Logs** with records containing accented user and item names.
2. Verify **Connections**, **Failed connections**, **Authentication lockouts**, **Errors**, **Copy**, **Administration**, **Items**, and **Knowledge base**.
3. Confirm that values stored as `é`, `&eacute;`, and `&amp;eacute;` all render as `é`.
4. Confirm that labels containing `<`, `>`, quotes, or entity-encoded markup are displayed as text and do not create DOM elements or execute scripts.
5. From **Authentication lockouts**, remove a lockout and confirm that the action still targets the exact stored login or IP value.

## Files changed

- `app/sources/main.functions.php`
- `app/sources/logs.datatables.php`
- `app/sources/kb.queries.php`
- `app/pages/utilities.logs.js.php`
- `tests/Unit/UtilitiesLogsEncodingTest.php`

No schema change or data migration is required.

### DIFF
--- a/app/includes/language/french.php
+++ b/app/includes/language/french.php
@@ -3012,7 +3012,7 @@ return array(
     'ops_hibp_occurrences' => 'occurrences HIBP',
     'ops_hibp_disabled' => 'La vérification Have I Been Pwned est désactivée. Les statuts déjà en cache restent affichés lorsqu’ils existent.',
     'ops_hibp_interval' => 'revérification tous les {days} jour(s)',
-    'ops_kpi_complexity_unknown' => 'Unknown complexity',
+    'ops_kpi_complexity_unknown' => 'Compléxité inconnue',
     'ops_kpi_created_period' => 'Créés sur la période',
     'ops_kpi_created_weak' => 'Créés faibles',
     'ops_activity_summary' => 'Synthèse de l’activité',
--- a/app/pages/utilities.logs.js.php
+++ b/app/pages/utilities.logs.js.php
@@ -350,13 +350,17 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
     }
 
     /**
-     * Escape a DataTables value before displaying it as HTML.
+     * Decode a legacy entity layer and escape the value before displaying it as HTML.
      *
      * @param {*} value Value to escape.
      * @return {string}
      */
     function escapeAuthenticationLockoutValue(value) {
-        return $('<div/>').text(value === null || value === undefined ? '' : value.toString()).html();
+        const normalizedValue = decodeHtmlEntities(
+            value === null || value === undefined ? '' : value.toString()
+        );
+
+        return $('<div/>').text(normalizedValue).html();
     }
 
     /**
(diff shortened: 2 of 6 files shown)


nilsteampassnet:
Merged in develop

PR #5317, opened 2026-08-03
Display user avatar in the topbar account menu

guerricv:
## Summary

This change displays the authenticated user's avatar in the account area at the top right of the TeamPass interface.

When no usable avatar is available, TeamPass keeps displaying the user's initials exactly as before.

The avatar is shown consistently in:

- the compact account chip in the topbar;
- the account header inside the dropdown menu.

## Current behavior

TeamPass 3.2.1.4 introduced an account chip containing the user's initials. Although the user's avatar and avatar thumbnail are already stored in the session, neither value was used by this new component.

As a result, users with an existing profile image still saw their initials in the topbar.

### Safe avatar resolution

`public/index.php` now resolves the account image from the existing session data:

1. use `user-avatar_thumb` when its file exists;
2. fall back to the full `user-avatar` file when needed;
3. keep the initials when neither file is usable.

The stored filename is reduced to its basename, checked with `is_file()`, and URL-encoded before being inserted into the generated URL.

No additional database query is performed.

### Account avatar rendering

Both account avatar locations retain the initials in their markup as the fallback content. When a valid image is available, it is placed over those initials.

The image:

- fills the existing circular account badge;
- uses `object-fit: cover` to preserve the badge proportions;
- works with both the regular and enlarged account badge sizes;
- is removed if loading fails, revealing the initials again.

### Live profile updates

The profile page now updates the topbar account avatar immediately after a successful avatar upload.

When the current avatar is deleted, the topbar image is removed immediately and the original initials become visible again. No page reload is required for either action.

## Security considerations

- The avatar filename is never used directly as a filesystem path.
- `basename()` prevents path components from being honored.
- `is_file()` prevents rendering stale database or session references.
- `rawurlencode()` and `htmlspecialchars()` protect the generated image URL.
- The image is decorative (`alt=""`) because the surrounding account control already has an accessible profile label and user identity.
- No user-supplied HTML is generated by the live profile update; the image element is built through jQuery's element API.

## Compatibility and impact

- Compatible with the existing TeamPass 3.2.1.4 session and avatar upload mechanisms.
- No database schema change.
- No install or upgrade script.
- No API change.
- No new language string.
- No new dependency.
- Existing initials behavior remains the fallback for all account types.

## Tests added

`tests/Unit/AccountAvatarDisplayTest.php` provides regression coverage for:

- thumbnail-first resolution with full-avatar fallback;
- safe filename and file-existence checks;
- presence of the avatar in both account locations;
- preservation of initials in both locations;
- circular cropped-image styling;
- immediate topbar updates after avatar upload and deletion.

## Suggested manual validation

1. Sign in with a user who has no avatar and confirm that initials are displayed in both account locations.
2. Sign in with a user who already has an avatar and confirm that the image is displayed in both locations.
3. Upload a new avatar from the profile page and confirm that the topbar changes without reloading the page.
4. Delete the avatar and confirm that the topbar immediately returns to the initials.
5. Confirm the result in light mode, dark mode, and the compact mobile layout.
6. Temporarily test a missing avatar file reference and confirm that initials are used instead of a broken image.

## Files changed

- `public/index.php`
- `public/assets/css/teampass.css`
- `app/pages/profile.js.php`
- `tests/Unit/AccountAvatarDisplayTest.php`

No data migration is required.

### DIFF
--- a/app/pages/profile.js.php
+++ b/app/pages/profile.js.php
@@ -103,6 +103,27 @@ if ($checkUserAccess->checkSession() === false || $checkUserAccess->userAccessPa
 
 
     // AVATAR IMPORT
+    const updateAccountAvatar = (avatarUrl) => {
+        $('.tp-account-avatar').each(function() {
+            const accountAvatar = $(this)
+            accountAvatar.find('.tp-account-avatar-img').remove()
+
+            if (typeof avatarUrl !== 'string' || avatarUrl === '') {
+                return
+            }
+
+            $('<img>', {
+                class: 'tp-account-avatar-img',
+                src: avatarUrl,
+                alt: ''
+            })
+                .on('error', function() {
+                    $(this).remove()
+                })
+                .appendTo(accountAvatar)
+        })
+    }
+
     // Both the button in the settings tab and the profile picture itself open the file dialog.
     // Plupload resolves browse_button through Dom.getAll(), so a list of ids is supported and
     // missing elements are simply ignored (profile edition disabled).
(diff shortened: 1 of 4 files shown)


guerricv:
redirecting to feature/lapr branch

PR #5322, opened 2026-08-05
Improve LAPR usability, item integration, and managed-account safety

guerricv:
## Summary

This pull request improves the day-to-day usability of Linux Account Password Rotation (LAPR), fixes the inline JavaScript rendering issue found while testing the managed accounts page, and integrates LAPR ownership directly into the TeamPass item experience.

The changes make LAPR easier to find, easier to configure with least privilege, and faster to operate after local Linux accounts have been discovered. They also prevent normal item edits from silently desynchronizing a managed Linux account and make LAPR relationships visible through contextual badges. They do not change the LAPR database schema or weaken any existing authorization, folder-access, host-key, or rotation safeguards.

## Context

During lab validation, the managed accounts page failed before initialization because adjacent PHP-generated JavaScript constants could be concatenated in the rendered response. Once that issue was corrected, several workflow improvements became apparent:

- administrators needed a quick way to filter the LAPR permission list;
- the long translated module label wrapped poorly in the sidebar;
- enrollment errors explained the missing capability but did not help an operator configure the endpoint;
- discovered Linux accounts could not be taken directly into the managed-account workflow;
- forced rotation still used the browser-native confirmation dialog instead of the TeamPass UI;
- the add-account form did not scale when the vault contained many items;
- re-adding a previously removed managed item caused a fatal duplicate-key exception because logical deletion retained the unique `item_id` value;
- LAPR-managed items were indistinguishable from ordinary items and could be edited in ways that desynchronized the vault from Linux.

## Integration dependencies and known overlaps

This working tree was prepared on top of changes from [PR #5320](https://github.com/nilsteampassnet/TeamPass/pull/5320) and [PR #5321](https://github.com/nilsteampassnet/TeamPass/pull/5321). Both PRs currently target `feature/lapr`. They do not overlap with each other, but this follow-up PR modifies files supplied by each of them.

| Prerequisite PR | Files changed by that PR | Overlap with this follow-up | Recommendation |
|---|---|---|---|
| PR #5320 — API personal-to-shared item move | `app/api/Model/ItemModel.php`, `app/sources/items.queries.php`, `app/sources/main.functions.php`, and `tests/Unit/Api/PersonalToSharedItemMoveTest.php` | `app/api/Model/ItemModel.php` and `app/sources/items.queries.php` | Integrate PR #5320 before this PR. Otherwise, rebase these two files manually and preserve both the personal-to-shared move implementation and the LAPR update/delete guards. |
| PR #5321 — user avatar in the top-bar account menu | `app/pages/profile.js.php`, `public/assets/css/teampass.css`, `public/index.php`, and `tests/Unit/AccountAvatarDisplayTest.php` | `public/index.php` | Integrate PR #5321 before this PR. The `public/index.php` used here starts from the PR #5321 version; without it, reapply the LAPR sidebar changes to the target version while preserving the avatar/top-bar changes as appropriate. |

PR #5320 and PR #5321 may be integrated in either order because their file lists are disjoint. The preferred sequence is therefore: merge both prerequisites, refresh/rebase this branch, resolve the three overlapping files above, then run the combined validation plan.

[PR #5316](https://github.com/nilsteampassnet/TeamPass/pull/5316) also touched `app/includes/language/french.php`. Its intended language change only translates `ops_kpi_complexity_unknown`; it does not intentionally remove Quick access strings. During the local integration of PR #5316, the Quick access block disappeared, most likely because the language file revisions were offset during conflict resolution. This follow-up restores that block explicitly, as described below.

### 1. Fix inline JavaScript rendering on the managed accounts page

- Serialize the LAPR session key and localization dictionary with `json_encode()` and the JSON hex flags appropriate for an inline script context.
- Terminate PHP-generated JavaScript constants explicitly with semicolons.
- Preserve valid JavaScript even when PHP consumes the newline immediately following a closing PHP tag.
- Apply the same safe JSON serialization approach to the updated LAPR endpoint and administration scripts.

This fixes the observed `Unexpected token 'const'` error and prevents translated apostrophes, quotes, or HTML-significant characters from breaking the script.

### 2. Add user search to the LAPR administration page

- Add a compact search field to the Permissions tab of `admin_lapr`, matching the existing API-user search presentation.
- Filter permission rows immediately by login or display name.
- Display the standard TeamPass empty-search result message when no user matches.
- Build user rows with jQuery text nodes and typed attributes instead of concatenating values into HTML.

### 3. Reorganize the LAPR sidebar and align operational permissions with item access

- Rename the module label to `LAPR` in `english.php` and `french.php` only.
- Preserve **Passwords** as the historical direct link to the TeamPass item vault for every non-admin user.
- For a non-admin user holding `can_manage_lapr`, render a separate expandable **LAPR** entry immediately below **Passwords**, without repeating the Passwords label inside another tree.
- Show the managed endpoint, managed account, and rotation policy links only while LAPR is enabled.
- For TeamPass administrators, keep only **LAPR settings** under **Operations** so the module can be enabled, configured, and delegated.
- Remove operational LAPR pages from the administrator allowlist in both source and Composer copies of `PerformChecks`.
- Update the central `laprCheckPermission()` gate so operational access requires an enabled module, a non-admin account, and the `can_manage_lapr` permission.
- Make the LAPR folder read/write helpers reject administrators as an additional item-access safeguard.

This role boundary matches the TeamPass design: administrators do not have item access, while every managed LAPR account depends on an existing TeamPass item.

### 4. Provide OS-aware endpoint enrollment remediation

- Parse `/etc/os-release` during the existing successful SSH capability test.
- Store the detected OS ID, compatibility identifiers, and normalized family in the endpoint's existing `os_info` JSON.
- Recognize Debian/Ubuntu, RHEL/Fedora, SUSE, Arch, Alpine, and a generic Linux fallback.
- When SSH succeeds but password rotation capability is missing, show:
  - the detected operating system;
  - prerequisite checks;
  - the appropriate package-manager commands;
  - a least-privilege `sudoers` rule limited to the resolved `chpasswd` executable;
  - `visudo` validation and the same no-op `sudo -n chpasswd` probe used by LAPR.
- When SSH fails before OS detection is possible, state that limitation and show only generic DNS, TCP, SSH service, account, and authentication diagnostics.
- Add a copy-to-clipboard action for the proposed commands.

The remediation deliberately does **not** request unrestricted sudo access. A non-root SSH account only needs passwordless execution of `chpasswd`, which is the exact privileged command used by the rotation path.

### 5. Add a direct managed-account action to discovery results

- Add a **Manage this account** action to each discovered Linux account.
- Carry the selected endpoint and discovered username into the existing add-account modal.
- Lock the endpoint selection to the endpoint that produced the discovery result.
- Request TeamPass items with an exact, parameterized login filter before presenting the searchable picker.
- Restrict the item selector to writable, active, non-personal, not-already-managed TeamPass items whose login exactly matches the discovered Linux account.
- Allow the operator to select the rotation policy and complete the existing managed-account creation flow.
- Disable saving and explain what to do when no matching TeamPass item exists.

LAPR still never creates vault items automatically. Password ownership remains explicit: the matching TeamPass item is the source of the Linux username and receives every newly rotated password.

### 6. Use the standard TeamPass rotation confirmation dialog

- Replace `window.confirm()` for forced password rotation with the shared `launchConfirmDialog()` helper.
- Use the standard TeamPass warning modal, translated title, translated message, standard cancel action, and explicit **Rotate now** action.
- Keep the existing background task, duplicate-task protection, polling, success handling, host-key mismatch handling, and manual-resynchronization warning unchanged.

### 7. Add a searchable managed-item picker

- Replace the preloaded item list in **Add a managed account** with the same server-side Select2 pattern used during endpoint enrollment.
- Search by item label or Linux login and show the folder path to disambiguate identical labels.
- Restrict results to writable folders, non-personal active items with a non-empty login, and items that are not already managed.
- Keep the discovery shortcut compatible with the picker by preselecting an exact-login match while allowing the operator to search other exact-login candidates.
- Bound each search response to 20 rows and keep all filters parameterized.

### 8. Safely reactivate logically deleted managed accounts

- Detect an existing `lapr_accounts` row for the selected TeamPass item before insertion.
- Return the existing functional “already managed” response when the relationship is still active, paused, or in error.
- When the row is logically deleted, reuse it instead of inserting a duplicate:
  - update the endpoint and cached Linux username;
  - apply the selected policy;
  - clear the previous rotation result, error, and retry state;
  - compute a fresh next-rotation date;
  - restore the account to `active`.
- Record whether the add audit event was a reactivation.
- Disable the modal save action while the request is in flight so a double click cannot submit two identical additions.
- Convert a concurrent unique-key race into a controlled functional conflict rather than an uncaught `MeekroDBException` and HTTP 500 response.

This directly fixes the lab failure `Duplicate entry '<item_id>' for key 'uniq_item_id'` without dropping the uniqueness guarantee or losing historical relationships.

### 9. Integrate LAPR roles into TeamPass items

- Add a batch relationship helper that identifies whether an item is:
  - the password target of a non-deleted managed account;
  - the SSH credential of one or more non-deleted endpoints;
  - or both.
- Display compact **LAPR managed** and **LAPR SSH credential** badges in folder item lists, global search results, and the item detail header.
- Colour the managed badge according to account state and expose richer endpoint, policy, next-rotation, scheduler, and endpoint-status context only to an authorized LAPR operator with write access to the item's folder.
- Lock the normal item `login` and password controls for managed targets while keeping labels, URLs, descriptions, tags, custom fields, and other metadata editable.
- Enforce the same managed-field ownership in the web update handler and REST API, returning a controlled error if a client attempts to bypass the UI.
- Keep credential-only items editable as an intentional recovery path, but warn that changing the stored secret does not change the remote Linux password and may break subsequent LAPR connections.
- Block item deletion through the web UI/backend and REST API while a managed account or enrolled endpoint still references it.
- Add **Rotate now** to a managed item's detail panel. It uses the existing standard TeamPass confirmation, LAPR background task, polling, SSH-first update, and failure handling; no second rotation implementation was introduced.
- Hide the legacy server-password action for LAPR-managed items so operators have a single unambiguous rotation path.

Direct item edits never launch SSH implicitly. LAPR rotation remains explicit or scheduler-driven, ensuring slow or failed remote work never runs inside an ordinary item-save request.

### 10. Restore Quick access language entries after integration

- Restore the `quick_access_*` menu strings in `english.php` and `french.php`:
  - title, recent, most-used, filter, and see-all labels;
  - empty-state labels for recent, frequent, and starred items;
  - no-match, ranking-hint, and usage-count labels.
- Restore `show_last_items` and `show_last_items_tip`, which label the setting controlling the Quick access panel.
- Keep the actual PR #5316 French translation of `ops_kpi_complexity_unknown` intact.

This is a local integration repair, not a claim that PR #5316 intentionally removed the Quick access feature or its translations.

## Security considerations

- No secret is added to UI responses, logs, OS metadata, or remediation commands.
- SSH failures still expose only the existing safe error taxonomy to the browser.
- OS-specific commands are chosen from a fixed family allowlist; untrusted remote OS text is escaped for display and is not copied into executable command lines.
- Hostnames and SSH usernames used in diagnostic examples are constrained before being inserted into the command block; unsafe values fall back to placeholders.
- The proposed sudo configuration grants only `NOPASSWD` access to the absolute `chpasswd` path and is validated with `visudo`.
- The direct discovery workflow does not bypass server-side folder write checks, item validation, username validation, endpoint status checks, or duplicate managed-item checks.
- Exact-login and interactive item searches remain parameterized, wildcard input is escaped, and search results are restricted to writable folders.
- Existing host-key verification and signed test-to-save snapshots are unchanged.
- Hiding administrator operational links is backed by `PerformChecks`, the central LAPR gate, and the folder-access helpers; direct URLs and AJAX calls are rejected as well.
- List and search responses receive only the LAPR booleans and managed-account status needed to render badges; endpoint names, hostnames, policies, and dates are returned only to an authorized LAPR operator with write access to the item folder.
- Managed login/password ownership and linked-item deletion rules are enforced server-side in both web and REST paths, not only by disabled controls or hidden buttons.
- Credential-only items remain editable deliberately: this preserves a recovery route if an endpoint password is changed out of band, while the warning makes the non-dynamic behaviour explicit.

## Database and compatibility

- No database migration is required.
- The existing unique constraint on `lapr_accounts.item_id` is preserved. Reactivation works with the current schema by updating the logically deleted row.
- Additional OS fields are stored in the existing `lapr_endpoints.os_info` JSON column.
- Previously enrolled endpoints remain compatible; endpoints tested after this change receive the normalized OS-family metadata.
- Language changes are intentionally limited to `english.php` and `french.php`. Other languages use TeamPass's existing English fallback for new detailed strings.

## Validation performed

- Confirmed manually in the lab that the managed accounts page initializes after the inline JavaScript rendering fix.
- Ran JavaScript syntax checks for:
  - `admin_lapr.js.php`;
  - `lapr_endpoints.js.php`;
  - `lapr_accounts.js.php`.
- Repeated those checks while simulating PHP's removal of the newline after `?>`, covering the original concatenated-constant failure mode.
- Verified that the forced-rotation path no longer calls `window.confirm(laprAccLang.rotateConfirm)`.
- Verified that the English and French LAPR module labels render as `LAPR` and that no other language file was changed for this label.
- Reviewed the endpoint remediation command builder for shell-value constraints and least-privilege output.
- Reviewed the discovery-to-item lookup to ensure the exact login filter is parameterized and existing server-side authorization remains authoritative.
- Reviewed the managed-item search to ensure it is bounded, parameterized, wildcard-safe, and scoped to writable folders.
- Traced the reported duplicate-key exception to the interaction between logical deletion and the existing `uniq_item_id` constraint, then verified that the add path now updates a deleted row and handles an insert race explicitly.
- Reviewed the modal event binding and added an in-flight submission guard so repeated clicks cannot create parallel add-account requests.
- Reviewed item list/search payloads to ensure only redacted LAPR badge metadata is exposed outside the authorized LAPR-operator detail path.
- Reviewed both web and REST item update/delete paths for managed-field and relationship guards.
- Confirmed that item-detail rotation reuses `start_rotation` and `rotation_status`; it does not perform SSH work in the item request thread.
- Corrected the missing closing parenthesis in the managed-item detail renderer and rechecked the surrounding LAPR JavaScript block for balanced parentheses.
- Compared the GitHub file lists for PR #5320, PR #5321, and PR #5316; confirmed that PR #5320 and PR #5321 do not overlap with each other and documented this PR's three prerequisite-file overlaps.
- Confirmed that PR #5316 changes only `ops_kpi_complexity_unknown` in `french.php`; the restored Quick access block is therefore documented as a local integration repair.
- Added unit coverage for the operational role boundary: delegated non-admin access, administrator rejection, disabled-module rejection, and administrator rejection by folder helpers.
- Reviewed the modified PHP signatures, return types, parameterized queries, and session-value casts for PHP 8.2 and the repository's PHPStan level 4 expectations.

PHP, Node.js, Composer, and a container runtime are not available in the local Windows workspace. PHP linting, JavaScript parsing, PHPStan, and automated tests for the final combined changes must therefore also run in the TeamPass runtime/container or CI environment.

## Suggested lab test plan

1. Open **Operations > LAPR settings** as an administrator and filter the permission list by login and name.
2. Disable LAPR and confirm that LAPR settings remain reachable while no operational LAPR menu is shown.
3. Enable LAPR and confirm that the administrator still sees only **LAPR settings**, then verify that direct requests to `lapr_endpoints`, `lapr_accounts`, and `lapr_policies` are rejected.
4. Sign in as a permitted non-administrator and confirm that **Passwords** still opens the item vault directly and that a separate **LAPR** entry appears immediately below it.
5. Test enrollment against an Ubuntu endpoint without the `NOPASSWD: chpasswd` rule:
   - confirm Ubuntu/Debian detection;
   - copy and run the proposed commands as an endpoint administrator;
   - rerun the test and confirm enrollment is accepted.
6. Test an invalid password or unreachable endpoint and confirm that generic diagnostics are displayed without claiming that the OS was detected.
7. Discover accounts on an endpoint and select **Manage this account**:
   - with one matching TeamPass item;
   - with multiple writable matching items;
   - with no matching item;
   - with an item that is already managed.
8. Complete a managed-account addition and verify the account appears in the main table with the selected endpoint and policy.
9. Remove that managed account, add the same TeamPass item again, and verify:
   - no HTTP 500 or `uniq_item_id` exception occurs;
   - the account reappears as active with a fresh next-rotation date;
   - a second simultaneous/non-deleted addition returns a normal “already managed” message.
10. Open **Add a managed account**, search by partial label and login, and verify the displayed folder paths and writable-folder scope.
11. Verify **LAPR managed** and **LAPR SSH credential** badges in a folder list, a global search result, and item details, including an item that has both roles.
12. Edit a managed item:
   - confirm that login/password controls are disabled;
   - confirm that changing another field still succeeds;
   - attempt a crafted web/API login or password update and confirm it is rejected;
   - confirm deletion is blocked until the managed account is removed.
13. Edit an SSH-credential-only item and confirm the warning is shown, recovery editing remains possible, and deletion is blocked while an endpoint references it.
14. Select **Rotate now** from both the managed accounts page and the managed item detail panel. Cancel once, then confirm once and verify the normal background rotation completes and the detail view refreshes.
15. Disable the scheduler and verify the item remains badged as managed while its detail panel reports that automatic scheduling is disabled.
16. Recheck browser developer tools and the PHP-FPM/Apache log for JavaScript errors, fatal errors, or unexpected LAPR requests on all modified pages.
17. Open the Quick access panel and its related setting in both English and French, and verify that no raw `quick_access_*`, `show_last_items`, or `show_last_items_tip` key is displayed.

## Deployment notes

- Deploy the modified PHP and PHP-generated JavaScript files together.
- Prefer integrating PR #5320 and PR #5321 before this PR, then rebase and review `app/api/Model/ItemModel.php`, `app/sources/items.queries.php`, and `public/index.php` carefully.
- If OPcache is configured without timestamp validation, reset OPcache or restart the relevant PHP-FPM service after deployment.
- Perform a hard browser refresh after deployment to avoid cached generated JavaScript.


guerricv:
As requested, I rewrote the source branch so this PR is now independent of #5320 and #5321, which are both still open and unmerged.

What was changed:
- kept the PR based directly on `feature/lapr` at `a0abb760`;
- removed the #5320 personal-to-shared item-move changes from `app/api/Model/ItemModel.php` and `app/sources/items.queries.php`;
- removed the #5321 avatar-related changes from `public/index.php`;
- preserved all LAPR-specific changes;
- folded the cleanup into the three affected commits, keeping the 22-commit structure without adding revert/fixup commits.

The branch was updated with `--force-with-lease`: `76f088a1` → `3110f2b5`.

Current PR diff: 22 files, 1,219 additions and 196 deletions. GitHub reports the PR as mergeable/clean, and `git diff --check` passes. PHP tests could not be run locally because no PHP runtime is available in the local environment.

guerricv:
Follow-up usability adjustments have been pushed in commit [25f074c](https://github.com/guerricv/TeamPass/commit/25f074c):

- Fixed the WebSocket max_connections retry loop by resetting the retry budget only after application-level authentication succeeds; the expected connection-limit condition is now logged as a warning. The source and public client copies remain identical.
- Reworked the LAPR indicators in the item list as icon-only markers next to the password-health indicator. The security shield is always placed to the left of the LAPR markers, while the full badges remain unchanged in the item detail view.
- Automatically focuses the Select2 search field when choosing the TeamPass credential item during endpoint enrollment.
- Replaced the native deletion confirmations for enrolled endpoints and managed accounts with the standard localized TeamPass warning modal.

The commit is a fast-forward on top of the cleaned PR head (3110f2b). Static diff/whitespace checks passed. Runtime and UI automation could not be run in the current workspace because PHP, Node.js, and the browser backend were unavailable.

PR #5333, opened 2026-08-14
Restore the item deletion confirmation button state

guerricv:
## Summary

Deleting an item from its detail card worked once per page load. After the first successful deletion, opening another item and attempting to delete it displayed the confirmation button in its previous disabled loading state, which prevented any further deletion from the detail card.

Deletion from the item list was not affected because that workflow uses a separate confirmation modal and button.

## Root cause

The detail-card deletion workflow disables `#form-item-delete-perform` and replaces its label with a loading spinner before sending the deletion request.

The button was restored when the server returned an error or when the AJAX request failed, but it was not restored after a successful deletion. The Bootstrap deletion modal is hidden and reused rather than destroyed, so its disabled button and spinner persisted when the modal was opened for the next item.

The same stale state could also occur when deletion was stopped by the client-side item-lock guard, because that branch returned before restoring the button.

## Implementation

- Restore the detail-card deletion button after a successful deletion, before hiding the modal.
- Restore the same button before returning from the item-lock guard.
- Reuse the existing localized `perform` label.
- Preserve the existing reset behavior for server-side errors and AJAX failures.

Together, these paths now restore the button for every handled terminal outcome:

- successful deletion;
- deletion rejected because the item is locked;
- server response reporting an error;
- failed AJAX request.

## User-visible behavior

- Users can delete multiple items successively from their detail cards without reloading the page.
- The confirmation button is enabled and displays its normal label whenever the deletion modal is reused.
- A lock-related rejection no longer leaves the confirmation button spinning and disabled.
- Deletion from the item list remains unchanged.

## Scope and compatibility

- Frontend-only change in `app/pages/items.js.php`.
- No change to the deletion endpoint or authorization checks.
- No database or installation change.
- No new dependency.
- No new or modified language string.
- No impact on PHPStan because the implementation only changes the embedded client-side behavior.

## Validation

- Confirmed that the button is disabled only when the detail-card deletion is launched.
- Confirmed that all handled completion paths now restore the button state.
- Confirmed that the successful path restores the button before the reusable modal is hidden.
- Confirmed that the item-list workflow still uses its separate warning modal and calls the same deletion function without changing its button behavior.

## Manual test plan

1. Open an item detail card and select **Delete item**.
2. Confirm the deletion and wait for the item card to close.
3. Open a second item detail card and select **Delete item**.
4. Confirm that the action button is enabled, displays the normal localized label, and can delete the second item.
5. Repeat with a third item without reloading the page.
6. Attempt to delete an item currently locked by another user and confirm that the button is restored after the warning.
7. Exercise a server-error or network-failure response and confirm that the button is restored.
8. Delete an item from the item list and confirm that the existing list workflow is unchanged.

## Risk assessment

Low. The change only restores the existing button state in two missing terminal branches. It does not alter deletion data, permissions, request payloads, server processing, or modal routing.

### DIFF
--- a/app/pages/items.js.php
+++ b/app/pages/items.js.php
@@ -1412,7 +1412,13 @@ $bip39Wordlist = loadBip39Wordlist($session->get('user-language') ?? 'english');
 
                 if (debugJavascript === true) console.info('SHOW DELETE ITEM');
                 if (store.get('teampassItem').user_can_modify === 1) {
-                    $('#modal-item-delete').modal('show');
+                    showItemDeleteModal(
+                        store.get('teampassItem').id,
+                        store.get('teampassItem').item_key !== undefined ? store.get('teampassItem').item_key : '',
+                        selectedFolderId,
+                        store.get('teampassItem').hasAccessLevel,
+                        true
+                    );
                 } else {
                     toastr.remove();
                     toastr.error(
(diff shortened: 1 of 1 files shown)


guerricv:
A follow-up UI consistency adjustment has been added to this PR.

The item list previously used the global `#warningModal`, while deletion from the item detail card used the dedicated `#modal-item-delete`. The dedicated modal follows the warning-styled destructive-action pattern already used elsewhere in TeamPass, with a centered dialog, warning header, explicit warning message, localized action label, and cancel action.

Both deletion entry points now prepare a shared deletion context and open `#modal-item-delete`. The common confirmation handler uses that context to delete the correct item and to preserve the existing difference between the two workflows: deleting from the detail card closes it, while deleting from the list does not.

This removes the list-specific temporary handlers attached to `#warningModal` and makes the confirmation UI consistent without changing the deletion endpoint, permissions, payload, or server-side processing.

nilsteampassnet:
Merged into 3.2.1.7

PR #5337, opened 2026-08-19
Modal hunt - Teampass 3.2.1.7

guerricv:
## Summary

This change removes the seven browser-native confirmation dialogs used by TeamPass application code and replaces them with the existing TeamPass confirmation UI. Destructive actions now use the shared `launchConfirmDialog()` helper, except for the project-integrity flow, which reuses its already-open TeamPass modal so that the integrity report remains available.

No backend operation or permission check has been changed. The destructive requests are still sent only after the user explicitly confirms the action.

### Profile

- Replaced the native confirmation shown before revoking a browser-extension token.
- Replaced the native confirmation shown before revoking an active API session.
- Kept the existing translated titles, messages, action labels, and cancellation label.
- Encoded the translated strings as JavaScript values before inserting them into the generated page script.

### Knowledge base

- Replaced native confirmations for deleting a knowledge-base entry, an attachment, and a comment.
- Reused the existing translated deletion messages.
- Added the existing translated cancellation label to the page translation catalog.
- Kept all AJAX success, error, edition-lock, and refresh behavior unchanged.

### Administration

- Replaced the native confirmation used to delete unknown files reported by the project-integrity check.
- Reused the currently open TeamPass modal instead of opening a nested Bootstrap modal.
- Restored the integrity report when the user cancels and kept the report visible while deletion runs.
- Namespaced and replaced the integrity action handlers to prevent duplicate callbacks after repeated checks.

### Tools

- Replaced the native confirmation used to delete a key-repair backup.
- Added `tools_delete_backup_confirm` to the English and French language catalogs.
- Reused existing translations for the selection warning, action label, cancellation label, and completion title.
- Preserved the selected backup identifier inside the confirmation callback and kept the delete button state synchronized with the request.

## Intentionally unchanged

- The `beforeunload` protection displayed during a database restore remains native. Browser navigation guards are controlled by the browser and are not equivalent to an application confirmation modal.
- Native dialogs contained in third-party compatibility or diagnostic paths remain unchanged, including DataTables, DataTables Scroller, and `download.js`.
- Inactive examples, comments, test payloads, and the local command-line `prompt()` helper are outside the scope of this UI change.

## Automated coverage

`NativeModalUsageTest` prevents native `alert()`, `confirm()`, and `prompt()` calls from returning to the four corrected page scripts. It also checks that the TeamPass modal integrations and the English/French backup-deletion translation exist, while explicitly preserving the database-restore navigation guard.

## Suggested manual verification

1. On the profile page, cancel and confirm both extension-token revocation and API-session revocation.
2. In the knowledge base, cancel and confirm deletion of an entry, an attachment, and a comment.
3. In the administration project-integrity dialog, cancel deletion and verify that the report returns; then confirm and verify that deletion results remain visible.
4. In the tools page, cancel and confirm deletion of a key-repair backup in both English and French.
5. Start a database restore and verify that the existing browser navigation warning remains unchanged.


nilsteampassnet:
Released in 3.2.2.0

PR #5340, opened 2026-08-21
LAPR account discovery and interface localization

guerricv:
## Summary

This change fixes the LAPR local-account discovery filter, removes untranslated text from the LAPR interfaces, applies TeamPass regional date/time settings to LAPR operational pages, and adds LAPR observability to the System Health and operational statistics pages.

The implementation deliberately keeps the preset labels stored in the database unchanged. Those English values remain stable internal identifiers, while localized display names are generated at the presentation boundary. User-created policy labels are always returned verbatim.

## System Health integration

System Health now includes a LAPR summary in the overview and a dedicated LAPR tab. The report is deliberately passive: loading the page never opens SSH connections or attempts to authenticate against managed endpoints.

The shared monitoring snapshot checks:

- whether the LAPR module and automatic scheduler are enabled;
- whether the next scheduler run or the oldest queued rotation task is overdue beyond a grace period;
- pending, running, and recently failed LAPR worker tasks;
- active, first-run-scheduled, retrying, overdue, error, and paused managed accounts;
- endpoint, managed-item, SSH-credential-item, and policy referential integrity;
- inactive, deleted, or personal items that cannot be used by the server-side rotator;
- managed-item login changes that no longer match the cached Linux username;
- disabled host-key verification and endpoints that no longer report rotation capability;
- duplicate endpoint targets and SSH credential items shared by several active endpoints.

The LAPR Health summary also reports the number of effective human operators. It follows the central permission gate: only enabled, non-deleted, non-administrator users holding `can_manage_lapr` are counted, service accounts are excluded, and the effective count is zero while the module is disabled. Permissions retained by disabled accounts are reported separately as a governance warning.

The grace period is the greater of ten minutes or two scheduler intervals. This prevents a rotation that has only just become due from being reported as overdue while the background handler is between normal runs.

Overdue and error accounts are critical findings. Intentionally paused accounts and rotations waiting for a scheduled retry are warnings, so planned maintenance remains visible without being reported as an immediate rotation failure. Endpoint-integrity findings are reflected in the global LAPR status even before the first managed account is attached.

Automatic rotation is optional. When its scheduler is intentionally disabled, System Health reports that fact as informational instead of treating it as a malfunction. A scheduler that is enabled but overdue, a stalled queue, or a failed worker remains an actionable health finding.

Recent rotation failures are returned as secret-free error codes and grouped into translated operational categories. Worker crashes are read from the background-task table as well as the LAPR audit log because an unexpected process failure can occur before a normal audit entry is written.

The LAPR section is part of the existing Health JSON report export. It contains only endpoint labels, Linux usernames, statuses, counts, dates, and stable error identifiers; it never selects or exports item passwords, SSH private keys, credential contents, or sharekeys.

## Operational statistics integration

The operational statistics dashboard now has a dedicated LAPR tab. Its selected period applies to completed rotation audits while account and endpoint status is explicitly presented as a current snapshot.

The tab provides:

- active versus total managed endpoints;
- compliant versus total managed accounts;
- rotation volume and success rate for the selected period;
- a stacked success/failure time series;
- the current managed-account state distribution;
- grouped failure causes;
- localized policy adoption;
- the endpoints with the most rotation failures, including their success rate and last failure date;
- a separate warning for unexpected failed worker tasks.

The existing personal-item and API filters are hidden while the LAPR tab is active because they do not apply to LAPR data. A retention warning is displayed when the requested period is longer than the configured LAPR audit retention.

All period queries use the indexed LAPR audit action/date columns and return aggregated rows. Raw audit history is not loaded by the statistics page.

## Rotation-result reliability

When the managed Linux account is also the endpoint SSH account but uses a separate TeamPass credential item, LAPR must update that credential item after changing the remote password. The return value of this second item update is now checked.

If synchronization fails, the rotation is no longer reported as successful. The managed account moves to the error state, both the credential-sync and rotation audit entries record `ERR_SSH_CREDENTIAL_SYNC_FAILED`, and the operator receives a dedicated manual-resynchronization message. This prevents the next SSH connection from silently relying on a stale credential item.

### Previous behavior

The discovery task read the endpoint account database through `getent passwd` and rejected only entries whose shell contained `nologin` or `/false`. The source comment stated that candidates must be `root` or have a UID greater than or equal to 1000, but the UID condition was not implemented.

Consequently, the standard Debian/Ubuntu `sync` account was displayed because it has UID 4 and `/bin/sync` as its shell.

### New behavior

The new `laprIsDiscoverableAccount()` helper applies all discovery requirements in one testable location:

- the entry must contain a valid username, numeric UID, and non-empty shell;
- `root` is accepted only when its UID is 0;
- non-root accounts require UID 1000 or greater;
- shells containing `nologin` or `/false` are rejected;
- the existing POSIX username validation remains mandatory.

`LAPRDiscoverTrait` now delegates candidate selection to this helper. The remote operation remains read-only and still uses `getent passwd` with the existing `/etc/passwd` fallback.

### Preset display names

The installer continues to seed these immutable database labels:

- `Standard (30 days)`
- `High Security (7 days)`
- `Weekly + rotate on enroll`

`laprPolicyDisplayName()` maps only recognized preset rows to language keys. The policies page, the managed-accounts table, the add-account modal, and the change-policy modal therefore receive localized preset names without a database migration.

Unknown future presets fall back to their stored label. Non-preset labels are never translated because they are user-provided data.

### Localized duration formatting

`laprPolicyOptionLabel()` formats policy selector entries with separate singular and plural language keys. Examples:

- English: `High security (7 days)` and `Daily (1 day)`
- French: `Haute sécurité (7 jours)` and `Quotidienne (1 jour)`

This also removes the previous hard-coded `d` suffix and avoids duplicating the duration already embedded in the stored preset label.

The French weekly preset uses the shorter `Hebdo avec rotation à l’ajout` display name. Preset labels and their badges are kept on one line, while custom labels retain the normal wrapping behavior.

### Managed-item detail panel

The **Managed by LAPR** panel on an item detail page now uses the same localized preset-name and duration helpers as the managed-accounts interface. Relation lookup returns the preset flag and rotation frequency as internal presentation metadata; the item-detail response converts recognized presets at its language-aware output boundary and removes that metadata before returning the response. Custom policy labels remain unchanged.

### DataTables controls

The endpoints, accounts, and policies tables now load the DataTables catalog shipped with the selected TeamPass language. This localizes the page-length selector, search label, result information, pagination controls, filtering messages, and processing state.

The implementation reads the catalog from the public language directory, falls back to the English catalog when the requested DataTables translation is unavailable, and overrides only the empty-table message with the LAPR-specific language key for each table.

### Regional date and time formatting

The managed-endpoints and managed-accounts handlers no longer expose SQL `Y-m-d H:i:s` values directly as display text. Endpoint checks, last and next rotations, and account-history events are formatted server-side with the configured TeamPass timezone, `date_format`, and `time_format` settings.

The background-task scheduler and worker now initialize the same configured timezone before comparing or writing LAPR datetimes. This makes future scheduler, rotation, retry, and audit timestamps consistent with timestamps created by web requests and prevents the PHP CLI timezone from leaking into the interface.

Each sortable table date also carries a Unix timestamp used only as its DataTables sort value. The interface therefore keeps chronological ordering independently of whether the selected display format is day-first, month-first, 12-hour, or 24-hour. Missing dates retain the existing neutral dash display.

### Global LAPR interface audit

All LAPR page templates, page scripts, and dedicated AJAX handlers were reviewed for visible hard-coded text. The audit additionally localized:

- endpoint and managed-account status badges;
- the endpoint host-key verification warning badge;
- endpoint remediation command comments while preserving executable shell commands verbatim;
- account-history event names, triggers, results, and range information;
- discovered-account table headings and its empty result message;
- endpoint modal accessibility text;
- endpoint/account validation errors previously returned as hard-coded English text or internal error identifiers.

Technical values such as SSH commands, operating-system names, database status identifiers, and diagnostic error codes remain unchanged. They are either not displayed as prose or must remain stable for troubleshooting and control flow.

### Other extracted UI text

The following policy/account strings now use the language catalogs:

- the empty policies table message;
- the empty managed-accounts table message;
- the empty managed-endpoints table message;
- the preset read-only indicator;
- invalid/not-found account, endpoint, policy, and task responses;
- unknown-action responses;
- modal close accessibility labels.

English entries provide the standard fallback for all languages. French translations were added for the reported interface.

## Tests

The automated coverage includes:

- acceptance of `root` with UID 0 and regular users with UID 1000 or greater;
- rejection of `sync`, reserved low-UID accounts, non-login shells, and a non-zero-UID `root` entry;
- English and French preset-name localization;
- singular and plural duration formatting;
- preservation of custom and unknown policy labels;
- use of the shared preset-name and localized-duration helpers by the managed-item detail response;
- use of the shipped French DataTables catalog by all three LAPR tables;
- the shortened French weekly-preset label and LAPR-specific empty-table messages;
- presence of every LAPR interface language key in the English and French catalogs;
- sentinels preventing the audited hard-coded table, history, endpoint, policy, and validation strings from returning;
- regional date/time formatting with day-first, month-first, 12-hour, and 24-hour settings;
- use of numeric timestamps for endpoint and account date-column ordering;
- explicit TeamPass-timezone initialization in both background-task entry points;
- pure account-health classification for healthy, first-run-scheduled, retrying, overdue, error, and paused states;
- the scheduler-derived overdue grace period;
- the intentional disabled-scheduler state remaining informational;
- effective operator access counts and dormant grants on disabled accounts;
- stable grouping of LAPR failure codes into presentation categories;
- Health report wiring, background-task coverage, and the absence of credential-secret selections;
- Statistics page wiring for snapshot and period metrics using the bundled Chart.js 2 API;
- a regression guard preventing SSH credential synchronization failures from being reported as successful rotations.

Relevant test files:

- `tests/Unit/LaprFunctionsTest.php`
- `tests/Unit/LaprLocalizationTest.php`
- `tests/Unit/LaprMonitoringLogicTest.php`
- `tests/Unit/HealthSystemConsistencyTest.php`
- `tests/Unit/OperationalStatisticsPageTest.php`

## Documentation

`docs/features/lapr.md` now states that discovery returns the real `root` account and regular accounts with UID 1000 or greater, while excluding reserved system accounts and non-login shells.

## Suggested validation

Run the focused test suite after installing development dependencies:

```bash
app/vendor/bin/phpunit tests/Unit/LaprFunctionsTest.php tests/Unit/LaprLocalizationTest.php
app/vendor/bin/phpunit tests/Unit/LaprMonitoringLogicTest.php tests/Unit/HealthSystemConsistencyTest.php tests/Unit/OperationalStatisticsPageTest.php
```

Recommended manual checks with the interface set to French:

1. Discover accounts on an Ubuntu endpoint and confirm that `root` and the UID 1000 user are shown while `sync` is absent.
2. Open **LAPR → Managed endpoints** and confirm that the DataTables controls, empty-table message, status badges, host-key warning, and endpoint modal are localized.
3. Test an endpoint without the required capabilities and confirm that the explanatory text and comments in the remediation command block are localized while the commands remain executable.
4. Open **LAPR → Managed accounts** and confirm that the DataTables controls, status badges, discovery results, and history modal are localized.
5. Open **LAPR → Rotation policies** and confirm that the same DataTables controls, preset names, empty-table message, and read-only indicator are localized.
6. Confirm that `Hebdo avec rotation à l’ajout` and its preset badge remain on one line.
7. Open the add-account and change-policy modals and confirm that preset names and durations use `jour`/`jours`.
8. Create a custom policy and confirm that its user-provided label is displayed unchanged.
9. Open an item managed by LAPR and confirm that its information panel displays a localized preset name and duration, for example `Haute sécurité (7 jours)` in French.
10. Open **System Health**, confirm that the overview LAPR badge matches the dedicated LAPR tab, and verify scheduler, account-state, action-item, and recent-failure sections.
11. Export the Health report and confirm that it contains the `lapr` section without item or SSH secrets.
12. Open **Statistics → LAPR**, switch between the supported periods, and verify the KPIs, charts, policy labels, and problematic-endpoint table.
13. Confirm that the personal-item and API filters disappear only while the LAPR statistics tab is active.
14. Force a controlled failed rotation and verify that both Health and Statistics show the translated failure category.
15. Grant LAPR access to enabled and disabled non-administrator users, then confirm that System Health counts only effective operators and reports dormant permissions separately. Disable the module and confirm that the effective operator count becomes zero.
16. Set TeamPass to a day-first 24-hour regional format and confirm that endpoint checks, account rotations, next rotations, and account-history dates follow it; repeat with a month-first 12-hour format and verify that table sorting remains chronological.

## Validation environment note

Static verification confirms that the English and French catalogs expose the same LAPR keys, all language keys referenced directly by the audited LAPR files exist in both catalogs, and none of the targeted hard-coded strings remain.

The current Windows workspace does not expose a PHP executable and contains production Composer dependencies only, without the PHPUnit binary. PHP linting and PHPUnit execution must therefore be completed in the project development container or CI environment.


guerricv:
### Authoritative module deactivation

Disabling LAPR now produces an effective operational stop instead of merely hiding access to the feature:

- System Health returns a neutral, complete disabled snapshot without querying or counting retained LAPR endpoints, accounts, permissions, queue entries, or failures.
- Statistics returns the same neutral snapshot plus an empty but structurally complete period payload. The LAPR tab remains visible for feature discovery, displays **Module disabled**, clears all KPIs and table rows, and destroys charts from a previous enabled response so no stale metric remains on screen.
- The Health LAPR tab also remains visible and replaces all operational values and detail tables with the neutral disabled message.
- Saving `lapr_enabled = 0` neutrally completes pending SSH-test, discovery, and rotation tasks with `ERR_LAPR_DISABLED` so they cannot execute after a later worker pickup.
- Every LAPR background task performs a fresh, direct database check of the master switch before dispatch. Rotation workers perform a second fresh check at the last safe point immediately before changing the remote password.
- If the SSH password change has already succeeded when the module is disabled, the worker intentionally completes the TeamPass item update. Stopping between the remote mutation and vault synchronization would create a more dangerous password mismatch.

No endpoint, account, policy, or audit record is deleted when the module is disabled. Re-enabling LAPR therefore restores the retained configuration without presenting it as active while the module is off.

### Endpoint and managed-account relationship guards

Server-side guards now enforce the supported relationship model both when the relationship is created and immediately before a rotation, protecting new and legacy data:

- the same normalized endpoint hostname/IP and SSH port cannot be enrolled twice;
- one endpoint may manage several TeamPass items only when they represent different Linux usernames;
- an endpoint/Linux-username pair can have only one non-deleted managed-account relationship;
- a TeamPass item can be the managed password for only one endpoint, preserving the existing unique-item invariant;
- a password-based SSH credential item cannot be reused by another endpoint, because rotating that connection account would synchronize only one remote password and invalidate access to the others;
- a private-key credential item may be deliberately shared by several key-authenticated endpoints;
- an SSH credential item cannot be mixed between password and key authentication modes;
- an item already used as a managed Linux password cannot later be selected as a new endpoint credential;
- a password credential may also be its endpoint's managed item only when it represents that same endpoint and the same SSH username;
- a private-key credential item can never be attached as a managed password.

Health System reports unsafe relationships retained from older data as actionable integrity findings: shared password credentials, duplicate managed endpoint/login pairs, and managed items reused as private-key credentials.

If a legacy duplicate endpoint or unsafe credential relationship reaches a worker, the pre-flight refusal is audited as a configuration failure and the managed account is placed in an error state immediately, including for scheduler-triggered work. This avoids an endless retry loop for a condition that requires an administrator to repair the relationship.

### Private-key credential rotation fix

The post-rotation SSH credential synchronization is now explicitly restricted to endpoints using password authentication. Key-authenticated endpoints still read their private key from the credential item's password field to connect, but the generated Linux password is never written back to that credential item.

This closes the critical case in which rotating the same Linux username used for a key-authenticated SSH connection could overwrite the stored private key and break all subsequent access. A pre-rotation relationship check also refuses legacy configurations in which a private-key credential item is itself registered as a managed password.

### TeamPass host self-management protection

LAPR now classifies targets that certainly or probably refer to the TeamPass host using local hostnames, loopback addresses, the server address, the configured TeamPass URL, and the current request host. DNS is deliberately not used because it is mutable, proxy-dependent, and unsuitable as an authorization boundary.

Self-management is handled as a break-glass workflow:

- it is blocked by default through the new `lapr_allow_self_management` administration setting, seeded to `0` for fresh installations and upgrades;
- when an administrator enables the setting, the endpoint enrollment modal shows a high-risk warning and requires a separate explicit acknowledgement before saving;
- the classification is persisted with the endpoint diagnostics and recomputed when practical, so renamed or legacy endpoints still receive protection;
- the scheduler excludes detected TeamPass endpoints;
- the rotation worker independently refuses every non-manual trigger for those endpoints;
- manual rotation remains available, but both the managed-accounts page and the managed-item detail panel show the stronger TeamPass-host warning in the interface and confirmation dialog;
- System Health reports a managed TeamPass endpoint as a warning to keep this exceptional dependency visible.

The detection is intentionally conservative. Environments where the public TeamPass hostname terminates on a separate reverse proxy can use the administrator override, but still receive acknowledgement, manual-only rotation, and Health warnings.

### Tests and validation added for this follow-up

The focused coverage now includes:

- pure self-target classification for loopback addresses, configured TeamPass URLs, and unrelated endpoints;
- neutral disabled Health and Statistics contracts;
- visible disabled-state Statistics behavior and stale-chart cleanup;
- cancellation and worker rejection of LAPR tasks after module deactivation;
- a last-moment master-switch check before the SSH password mutation;
- endpoint uniqueness and credential-role guard wiring;
- managed endpoint/login uniqueness and legacy pre-rotation relationship checks;
- the password-authentication condition around SSH credential item synchronization;
- default-off installation and upgrade seeding of the self-management override;
- enrollment acknowledgement, scheduler exclusion, worker manual-only enforcement, and item-detail warnings for TeamPass self-management.

Relevant test files for this follow-up:

- `tests/Unit/LaprFunctionsTest.php`
- `tests/Unit/LaprMonitoringLogicTest.php`
- `tests/Unit/LaprSafetyWiringTest.php`
- `tests/Unit/OperationalStatisticsPageTest.php`

Recommended manual regression checks:

1. Enable LAPR, enroll endpoints and managed accounts, then disable it. Confirm that Health and Statistics keep their LAPR tabs visible but show only the neutral module-disabled state, with no previous KPI, chart, table, operator, or queue values.
2. Queue an SSH test, discovery, or rotation and disable LAPR before worker pickup. Confirm that the task ends with `ERR_LAPR_DISABLED` and performs no SSH mutation.
3. Add two different Linux usernames to one endpoint and confirm both are accepted. Attempt to attach a second item with the same Linux username to that endpoint and confirm it is refused.
4. Attempt to reuse one password credential item across two endpoints and confirm it is refused. Repeat with one private-key credential item across two key-authenticated endpoints and confirm it is accepted.
5. Attempt to manage an item used as a private-key credential and confirm it is refused. Rotate a different managed account on a key-authenticated endpoint and confirm the private-key item remains byte-for-byte unchanged.
6. Attempt to enroll the current TeamPass host with the override disabled and confirm it is refused. Enable the override, repeat the test, acknowledge the warning, and confirm that enrollment succeeds only after acknowledgement.
7. Confirm that the detected TeamPass endpoint appears in Health, is skipped by the automatic scheduler, and presents the strong warning before a manual rotation from both the managed-accounts page and the item detail panel.

The current Windows workspace still has no PHP or Node.js executable and no PHPUnit binary. Static source verification was completed locally; PHP linting, PHPUnit, PHPStan, and browser-level regression checks must run in the project development container or CI environment.

### Modal consistency audit

All dedicated LAPR interfaces, the LAPR sections of Statistics and System Health, and the managed-item integration were audited for native browser dialogs. No use of `alert()`, `confirm()`, `prompt()`, or their `window.*` variants remains in the LAPR user journeys.

Endpoint enrollment, account management, discovery, history, and policy editing use the Bootstrap modal components rendered by TeamPass. Rotation and deletion confirmations use the shared TeamPass `launchConfirmDialog()` helper. Non-blocking status and error feedback continues to use the standard TeamPass toastr notifications.

`tests/Unit/LaprSafetyWiringTest.php` now includes a regression guard that scans every LAPR page script and the complete managed-item LAPR block for native browser dialogs. It also verifies that TeamPass confirmation dialogs and Bootstrap modals remain wired into these workflows.

### Rotate-on-enrollment execution fix

The `rotate_on_enroll` policy flag was stored and displayed but was not consumed when a managed account was created. Account creation only read the policy frequency, scheduled the first due date one full period later, and returned without creating a background task. The weekly preset therefore behaved like an ordinary seven-day policy instead of rotating immediately.

Managed-account creation now reads both `frequency_days` and `rotate_on_enroll`. When the flag is enabled, it immediately queues a deduplicated `lapr_rotation` background task with the existing `enroll` trigger, wakes the TeamPass background handler, and returns the task identifier to the interface. The account modal closes normally while the managed-accounts page polls the task and reports the final rotation result using the existing LAPR progress notifications.

This first rotation is asynchronous but independent of the automatic scheduler switch: the background-task infrastructure must be operational, while `lapr_scheduler_enabled` does not need to be enabled. On success, the worker computes the following rotation from the actual completion time. The TeamPass-host break-glass rule remains authoritative: self-target endpoints do not receive an enrollment-triggered task and the interface explains that their rotations are manual-only.

The enrollment trigger is also localized in account history. Focused coverage now verifies the queue decision, self-target exclusion, policy-flag lookup, task payload and deduplication wiring, background-handler wake-up, interface polling, and translated history trigger.

#### Reactivation follow-up and HTTP 500 fix

The first implementation of the deduplication lookup incorrectly selected `id` from `background_tasks`; the actual primary key is `increment_id`. Policies without enrollment rotation did not execute that lookup, while the weekly preset did, producing an SQL “unknown column” exception and HTTP 500 immediately after submission. The lookup now consistently selects and orders by `increment_id`, matching every existing task-status endpoint and the table schema. Account creation/reactivation, its audit entry, and the optional enrollment task now share one database transaction, so any future queueing error rolls the relationship change back instead of leaving an active account without its requested first rotation.

The remove/re-add lifecycle was hardened at the same time. Removing a managed account now neutrally completes every pending rotation for that account with `ERR_ACCOUNT_DELETED`. A worker that was already connecting re-reads the relationship state immediately before `changePassword()` and stops if the account was removed. Every worker-side account-state update also excludes rows whose status is `deleted`, preventing a late success, failure, or retry update from resurrecting a removed relationship. Re-adding the item therefore starts from the reset account row and, when requested by the policy, creates or follows only a currently executable enrollment task.

The static regression coverage now checks the real `background_tasks.increment_id` key, cancellation of pending rotations, the last-moment relationship check before the SSH mutation, and the non-resurrection conditions on all worker account updates.

guerricv:
### Automatic LAPR scheduler settings fix

Database verification after an overdue-rotation test showed that all LAPR configuration rows were correctly stored in `teampass_misc` under `type = 'admin'`, but the background handler's generic setting helpers always queried `type = 'settings'`. Consequently, the automatic scheduler always received the disabled defaults, never created a task with the `scheduler` trigger, and left `lapr_scheduler_next_run_at` at `0`. The System Health overdue finding was therefore accurate: the clock change exposed an automatic-scheduling defect rather than a Health calculation error.

The setting helpers now accept an explicit namespace while preserving `settings` as the default for the existing backup and security schedulers. Every LAPR scheduler read uses the `admin` namespace for the module switch, scheduler switch, interval, and next-run timestamp. Initializing or advancing the next-run timestamp writes back to that same `admin` row and invokes the standard TeamPass configuration-cache invalidation. LAPR audit cleanup now also reads `lapr_audit_retention_days` from `admin`, so a configured retention period is no longer silently replaced by the 365-day fallback.

No database migration or duplicate `settings` rows are required. Existing installations already hold the correct `admin` rows; deploying the PHP change is sufficient. When `lapr_scheduler_next_run_at` is `0`, the first background-handler pass initializes it to the current time plus the configured interval, and a later pass evaluates due accounts. Focused static regression coverage verifies all LAPR namespace reads, the scoped helper signatures, same-namespace next-run writes, and configuration-cache invalidation.

For manual validation, keep all operating-system clocks synchronized. Set one test account's `next_rotation_at` to the past and set the existing `admin/lapr_scheduler_next_run_at` value to a past Unix timestamp, then let the normal background handler run. A new `lapr_rotation` task with `"trigger":"scheduler"` should be created and completed; the account's next rotation should then be recomputed from the policy. This avoids the unrelated side effects caused by advancing domain and application clocks by several days.

guerricv:
### LAPR system actor in TeamPass item history

Automatic LAPR rotations write the item password-history event with TeamPass's internal `TP_USER_ID`, whose intentionally empty first and last name caused the item history card to render a blank author above “Password updated”. Manual and enrollment-triggered rotations already retain the initiating user's identifier and were not affected.

The item-history query now includes the log actor identifier and recognizes password-change entries owned by `TP_USER_ID` as automatic LAPR activity. Their login and display name are replaced at response time with the existing localized LAPR system label (`LAPR system` / `Système LAPR`). The database journal remains unchanged, real user attribution is preserved for user-triggered rotations, and previously recorded automatic rotations are fixed immediately without a migration or data rewrite.

Focused static coverage isolates the standard item-history endpoint and verifies that it selects the actor identifier, restricts the substitution to the internal TeamPass user and password-change action, and obtains the displayed label from the language catalogue rather than hardcoding interface text.

nilsteampassnet:
Delivered in 3.2.2.0

PR #5343, opened 2026-08-24
Preserve API password history and expose item revision timestamps

guerricv:
## Description

This PR aligns API password updates with the TeamPass web application and exposes a durable date
for the latest functional item revision. It addresses two requirements from the mobile client:

1. Preserve the former password in TeamPass history when the password is changed through the API.
2. Return the exact date associated with the current item revision so clients can display the last
   functional modification date.

### Encrypted password history for API updates

When `PUT /item/update` receives a non-empty password, the server now:

1. Evaluates the optional optimistic-concurrency `revision` precondition before password work.
2. Recovers the current object key through `decryptUserObjectKeyWithMigration()`, preserving the
   transparent phpseclib v1-to-v3 migration behavior.
3. Decrypts the current password with `teampassDecryptPasswordValue()`.
4. Compares the current and submitted passwords with `hash_equals()`.
5. Encrypts the previous cleartext password with the TeamPass application key before mutating the
   item.
6. Updates the password and its sharekeys through the existing API flow.
7. Adds an `at_modification` / `at_pw` entry to `log_items`, with the encrypted previous password
   in `old_value` and `TP_ENCRYPTION_NAME` as its encryption type.

If the current password or its key cannot be recovered, the request fails with `422` before the
item is changed. Failure to prepare the encrypted history value returns a generic `500` response
without exposing cryptographic details.

Submitting the current password is a password no-op. By itself, it does not rewrite the
ciphertext, redistribute sharekeys, add a history entry, or allocate a new revision.

### Durable item revision timestamp

The items table receives a nullable `BIGINT UNSIGNED` column named `revision_changed_at`. This is
the Unix UTC timestamp paired with the current `revision`; it is not a general row update date.

`bumpItemRevision()` computes one timestamp and stores it in both:

- `items_revisions.changed_at`; and
- `items.revision_changed_at`, alongside `items.revision`.

The revision and timestamp therefore remain one logical pair. Reads, access logs and
ciphertext-only migrations continue not to allocate functional revisions.

`revision_changed_at` is returned by:

- `GET /item/get`
- `GET /item/inFolders`
- `GET /item/findByUrl`
- `GET /item/changes`, for changed items and removed-item tombstones
- `POST /item/create`
- `PUT /item/update`

The delta feed takes both values from the winning journal entry after per-item deduplication. This
prevents a scanned revision from being paired with the timestamp of a newer concurrent item state.
Tombstone timestamps also come from the journal because a purged item no longer has an item row.

### Install, upgrade and compatibility

- Fresh installations create `items.revision_changed_at` in
  `public/install/install-steps/run.step5.php`.
- Upgrades add the column idempotently in `public/install/upgrade_run_3.2.2.php`.
- The upgrade backfills a date only when the exact current `(item_id, revision)` journal row still
  exists. Revision `0` and revisions whose journal entry was pruned remain `NULL`.
- The migration deliberately does not fall back to `items.updated_at`, whose semantics are not
  reliable for a functional content revision.
- The API change is additive. Existing clients may ignore the new field; clients that consume it
  must accept `null` when an exact date is unavailable.
- The normal TeamPass install/upgrade process must run before the updated API serves item queries,
  because those queries select the new column.

### Atomicity

This PR keeps the existing TeamPass API write model. It does not introduce a new transaction
across the item row, sharekey distribution, audit logging, cache refreshes and WebSocket side
effects.

Within that model, the revision precondition, current-password recovery and history encryption all
run before the item mutation. The existing item-write, dependent-sharekey and audit/cache ordering
is retained. A broader all-or-nothing transaction would require a separate refactor of helpers that
perform multiple writes and intentionally isolate audit failures from the main operation.

The OpenAPI 3.1 contract and the API and architecture documentation are updated accordingly.

## Related issue

No related issue number was provided.

## Type of change

- [x] Documentation
- [ ] Translation
- [ ] Refactor / maintenance

## How has this been tested?

The following local static checks were completed successfully:

- `app/api/openapi.json` parses successfully with PowerShell `ConvertFrom-Json`.
- The OpenAPI schemas expose `revision_changed_at` for full items, item summaries, create/update
  results and removed-item tombstones.
- Source guards confirm that the revision precondition and encrypted-history preparation occur
  before the item write, and that the password audit entry occurs after sharekey work.
- Both fresh-install and upgrade definitions contain the new column, and the upgrade uses the
  exact `(item_id, revision)` journal join for its backfill.
- The 15 files published on `guerricv/TeamPass:api_update` exactly match the reviewed local files.
- The branch is 15 commits ahead and 0 commits behind `nilsteampassnet/TeamPass:develop`.
- No conflict markers, `var_dump()` calls or `console.log()` calls were found in the PR files.

Tests added or updated:

- `ItemUpdateLogicTest` covers unchanged and changed passwords and the timing-safe comparison.
- `ItemPasswordHistoryTest` guards precondition ordering, migration-aware recovery, fail-closed
  behavior, history preparation, audit ordering and secret-free error logging.
- `ItemRevisionsLogicTest` covers revision/timestamp normalization and delta deduplication.
- `ItemRevisionTimestampTest` covers both schema paths, exact backfill, paired timestamp writes,
  OpenAPI coverage, `findByUrl`, delta items and tombstones.

PHP syntax checks, PHPUnit, PHPStan and database-backed install/upgrade scenarios were not executed
locally because this workspace has no PHP CLI, WSL installation or test database. GitHub currently
reports no check runs for the branch head. Opening the PR against `develop` should trigger the
repository's Quality workflow (PHPUnit and PHPStan) and CodeQL workflow.

No runtime role matrix has therefore been exercised yet: admin, manager, standard and read-only
users are untested, with personal folders both enabled and disabled. The following scenarios should
be covered before merge:

1. Change an item password through the API and confirm that the former password appears in the web
   history as an encrypted `at_pw` value.
2. Submit the same password alone and confirm that the revision, timestamp, ciphertext, sharekeys
   and history remain unchanged.
3. Verify item reads and delta synchronization with public and personal folders.
4. Verify a fresh installation and an upgrade with revision `0`, an exact journal match and a
   pruned journal entry.
5. Verify changed-item and delete/purge tombstone revision/timestamp pairs.

## Checklist

- [ ] The test suite passes (`php app/vendor/phpunit/phpunit/phpunit`)
  (not applicable: no query handler was added)
  (not applicable: no `teampassclasses` file was changed)
  (no `app/vendor/` file is present in the branch diff)

- [ ] No schema change

Schema changes are included in both `public/install/upgrade_run_3.2.2.php` and the fresh-install
path. The checkbox remains open because the fresh installation and upgrade were not executed
against a database in this workspace.

## Screenshots

Not applicable. This PR changes the API, database schema and backend behavior without adding or
changing a web UI.

### DIFF
--- a/.claude/docs/api-reference.md
+++ b/.claude/docs/api-reference.md
@@ -151,9 +151,14 @@ Get item(s) by ID or label.
 
 **Pagination:** label/description searches return `X-Total-Count` (total matches in accessible folders, before per-item sharekey filtering).
 
-**Response:** array of item objects `{ id, revision, label, description, login, email, url, password, path, folder_id, folder_label, has_otp, favicon_url, tags, fields }`.
+**Response:** array of item objects `{ id, revision, revision_changed_at, label, description, login, email, url, password, path, folder_id, folder_label, has_otp, favicon_url, tags, fields }`.
 
 **`revision`** — monotonic item revision, allocated from the `teampass_items_revisions` journal on every content change (item row, custom fields, tags, attachments, OTP, move, delete, restore). `0` = never changed since the column was introduced. Also returned by `item/inFolders`, `item/findByUrl`, `item/create` and `item/update`. See `architecture-item-revisions.md`.
+
+**`revision_changed_at`** — Unix UTC timestamp in seconds paired with `revision`. It changes only
+when a functional content revision is allocated, never on a read or ciphertext-only rewrite.
+`null` means the exact date is not known, notably for revision 0 or when an upgrade could not find
+the current revision in the prunable journal. Returned everywhere `revision` is returned.
 
 **Custom fields:** `fields` is an array of `{ id, title, type, masked, value }` for the item's folder-associated categories. Encrypted values are decrypted via `decryptUserObjectKeyWithMigration()` on `sharekeys_fields` (+ `base64_decode`); empty when no sharekey is available yet. Only present when `item_extra_fields` is enabled. Also returned by `item/inFolders`.
 
@@ -181,7 +186,7 @@ Find items by URL match.
 
 **Params:** `url` (string). The `%` and `_` characters are escaped before the LIKE query.
 
-**Response:** array of `{ id, label, login, url, folder_id, has_otp, favicon_url }`. Empty result → `200` + `[]`.
+**Response:** array of `{ id, revision, revision_changed_at, label, login, url, folder_id, has_otp, favicon_url }`. Empty result → `200` + `[]`.
 
 **Permissions:** `allowed_to_read`.
 
@@ -210,7 +215,7 @@ Delta feed for offline clients (mobile vault). Answers "what must I apply since 
 
 **Params:** `since` (int, **required**, exclusive; `0` on a first sync), `limit` (default 200, max 1000).
 
-**Response:** `{ cursor, has_more, full_sync_required, changed[], removed[] }`. `changed` carries full item payloads (same shape and same code path as `item/get`); `removed` is `{ id, revision, reason }` with `reason` in `deleted` | `purged` | `out_of_scope`.
+**Response:** `{ cursor, has_more, full_sync_required, changed[], removed[] }`. `changed` carries full item payloads (same shape and same code path as `item/get`); `removed` is `{ id, revision, revision_changed_at, reason }` with `reason` in `deleted` | `purged` | `out_of_scope`. The revision/date pair comes from the winning journal row after deduplication.
 
 **Cursor-based, not offset-based** — a change feed has no stable total, so no `X-Total-Count`. Store `cursor`, resend it as `since`, repeat while `has_more`.
 
@@ -244,7 +249,7 @@ Create a new item.
 
 **Custom fields:** `fields` = array of `{ id, value }` (field id + value). Encrypt-before-INSERT for encrypted categories; creator sharekey created synchronously, other users via the `new_item` background task. Only fields tied to the folder are stored; empty values ignored. Requires `item_extra_fields`.
 
-**Response 201:** `{ error: false, message, newId }` + `Location: /api/v1/item/get?id=<newId>` (path-absolute reference). Validation failures → `422`; missing fields → `400`; folder not allowed / read-only → `403`.
+**Response 201:** `{ error: false, message, newId, revision, revision_changed_at }` + `Location: /api/v1/item/get?id=<newId>` (path-absolute reference). Validation failures → `422`; missing fields → `400`; folder not allowed / read-only → `403`.
 
 **Permissions:** `allowed_to_create`. Blocked with 403 if folder is read-only for user.
 
@@ -257,6 +262,12 @@ Update an existing item. **Only PUT is accepted** — POST returns 405.
 **Body:** `id` (required), at least one of: `label`, `password`, `description`, `login`, `email`, `url`, `tags`, `anyone_can_modify`, `icon`, `folder_id`, `totp`, `fields`.
 
 **Optimistic concurrency — `revision` (optional).** The revision the client's edit was based on. When it differs from `items.revision`, the update is rejected with `409` and **nothing is written**; omitting it keeps last-writer-wins, so existing clients are unaffected. It is a **precondition, not an updatable field**: it is absent from the `$updateableFields` list in `ItemController::updateAction()`, so `{id, revision}` alone still answers `400 'At least one supported field to update must be provided.'`, and it never triggers the personal→shared move conflict guard.
+
+**Password history.** When a non-empty `password` is supplied, the server decrypts the current
+value before any mutation and compares it with `hash_equals()`. An unchanged password is a password
+no-op: by itself it causes no ciphertext rewrite, sharekey fan-out, password-history row or revision. A real change stores the
+previous value in `log_items.old_value`, encrypted with the application master key and tagged
+`at_pw`, exactly like the web UI. Failure to recover or prepare the old value aborts before mutation.
 
 **Custom fields:** `fields` = array of `{ id, value }`. Created if absent, updated only when the value changed (current value decrypted for comparison); encrypted fields re-encrypted and sharekeys refreshed synchronously for all eligible users (consistent with the password path). Empty values ignored. Requires `item_extra_fields`.
 
--- a/.claude/docs/api-reference.md
+++ b/.claude/docs/api-reference.md
@@ -247,9 +247,18 @@ Create a new item.
 
 **Body:** `label`, `password`, `folder_id`, optional `description`, `login`, `email`, `url`, `tags`, `totp`, `fields`.
 
+**Optional idempotency:** `Idempotency-Key: <opaque-key>` (1–128 visible ASCII characters,
+without spaces). Its identity is scoped to the authenticated user and `item.create`. The first
+accepted request returns the normal `201`; an identical replay within 90 days returns the exact
+stored result and `Idempotency-Replayed: true` without repeating any write or side effect. Reusing
+the key with a different functional payload returns `409`. A concurrent request still holding the
+key's processing lease also returns `409` with `Retry-After`. The fingerprint covers every
+functional create field, including credential, TOTP and custom-field values, but is a
+server-secret HMAC: neither those values, the request body nor the raw key is persisted.
+
 **Custom fields:** `fields` = array of `{ id, value }` (field id + value). Encrypt-before-INSERT for encrypted categories; creator sharekey created synchronously, other users via the `new_item` background task. Only fields tied to the folder are stored; empty values ignored. Requires `item_extra_fields`.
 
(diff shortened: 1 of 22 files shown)


nilsteampassnet:
Thanks for this PR — the write-up is thorough and the design is sound. About the failing CI: there are **two distinct causes, and only one of them is yours.**

### 1. The nine `Constant SECUREFILE not found.` errors are not from your PR

The `Quality` workflow is recent (added 2026-08-20) and had **never** passed — it failed identically on `develop` and `master`, on every run since it was created.

The `PHPStan (level 4)` job bootstraps a placeholder `app/config/settings.php` copied from `app/config/settings.sample.php`, because the real file holds the database credentials and is never committed. That sample was out of sync with what the installer actually writes: it declared every constant of a generated `settings.php` **except `SECUREFILE`** (which `run.step6.php:445` and `tp.functions.php:109` do write).

Fixed on our side in #5345, now merged into `develop`. Nothing for you to do beyond picking up `develop`.

### 2. One error is yours — `app/api/Model/ItemModel.php:2010`

```
Strict comparison using !== between non-falsy-string and null will always evaluate to true.
```

```php
if ($passwordChanged === true && $encryptedPreviousPassword !== null) {
```

`$encryptedPreviousPassword` is assigned in the same block as `$passwordChanged = true` (lines 1831 and 1849), and the `RuntimeException` just above already guarantees it is not empty. PHPStan's conditional narrowing therefore proves the second clause is always true on that branch — the guard is dead, not wrong. Replace it with:

```php
if ($passwordChanged === true) {
```

and pass `(string) $encryptedPreviousPassword` to `logItems()`. Behaviour is strictly unchanged.

### Updating your branch

```bash
git remote add upstream https://github.com/nilsteampassnet/TeamPass.git   # if not already set
git fetch upstream
git checkout api_update
git merge upstream/develop          # picks up the SECUREFILE fix
# apply the ItemModel.php:2010 change
git commit -am "Drop the redundant null guard on the password history value"
git push origin api_update
```

The push matters: GitHub does **not** re-run a PR's checks when its base branch advances on its own. Your merge of `develop` and the `ItemModel.php` fix ship together in one push.

### Running the checks locally

You mentioned having no PHP CLI — a `php:8.3-cli` container is enough:

```bash
rm -rf app/vendor/phpstan app/vendor/phpunit app/vendor/symfony/cache
composer install
cp app/config/settings.sample.php app/config/settings.php   # if you have no instance
php app/vendor/bin/phpstan analyse --memory-limit=2G
php app/vendor/phpunit/phpunit/phpunit
git checkout -- app/vendor/composer/                        # LAST, and not optional
```

That final `git checkout` is mandatory: `composer install` rewrites `app/vendor/composer/` into its *dev* form, and shipping that autoloader fatals every installation. The `Committed autoloader is production-only` job catches it — it is green on your PR today, so just don't break it while running the tools.

Both test jobs (PHP 8.2 and 8.3) already pass, and the rest of the PR is not implicated.

guerricv:
Implemented the requested follow-up:

- merged the latest upstream `develop`, which includes #5345 and its `SECUREFILE` sample-configuration fix;
- removed the redundant `$encryptedPreviousPassword !== null` guard reported by PHPStan;
- kept the `$passwordChanged === true` condition and explicitly cast the history value to `string` when calling `logItems()`;
- added regression assertions covering the simplified condition and explicit cast.

The behavior is unchanged: `$encryptedPreviousPassword` is assigned a non-empty string in the same branch that sets `$passwordChanged = true`, and failures abort before the audit call is reached.

The updated branch has been pushed. All refreshed GitHub Actions checks pass: PHPStan level 4, PHP 8.2 and 8.3 tests, the committed-autoloader check, and CodeQL. Scrutinizer has also been re-triggered and is still running at the time of writing.

guerricv:
# follow-up: idempotent create/delete and conditional delete

This follow-up extends PR #5343 for future bidirectional offline synchronization. No mobile or
browser-extension code is changed.

## API contract added

- `POST /item/create` accepts an optional `Idempotency-Key` (1–128 visible ASCII characters,
  without spaces).
- The first keyed create keeps the normal `201`, `Location`, `newId`, `revision` and
  `revision_changed_at` response.
- Replaying the same user/key/canonical intent returns the stored response with
  `Idempotency-Replayed: true` and performs no write or side effect.
- Reusing the key with another payload returns `409`; an identical request still processing
  returns `409` and `Retry-After`.
- `DELETE /item/delete` accepts an optional unsigned `revision` query precondition. A stale value
  returns `409` before mutation; omission preserves the legacy behavior.
- Successful delete responses now also contain the exact deletion `revision` and
  `revision_changed_at` pair visible in `GET /item/changes`.
- Delete accepts the same optional idempotency header. Its intent covers item id and expected
  revision; replay never deletes again, including after a later restore.

Requests without the new header, and deletes without `revision`, remain backward compatible.

## Persistence and security

- Added the prefixed `api_idempotency` table with a unique
  `(user_id, operation, key_hash)` identity, request fingerprint, processing lease, replay-safe
  result and lifecycle timestamps.
- Added nullable unique `items.api_idempotency_id`, linking a keyed create to its reservation so a
  stale processing state can recover a committed item without executing its side effects again.
- Raw keys, JSON bodies, passwords, TOTP secrets and custom-field values are never persisted.
  Key and intent hashes use HMAC-SHA-256 with a domain-separated key derived from the TeamPass
  instance secret.
- Create fingerprints cover every functional field and recursively canonicalize associative JSON
  properties, while preserving meaningful list order.
- Records are retained for a 90-day replay window. The existing orphan-maintenance task clears
  expired item links and records, but never removes an active processing lease.

Both the fresh-install path and `upgrade_run_3.2.2.php` contain the schema changes and indexes.
The upgrade remains idempotent.

## Atomicity and crash handling

- The durable processing reservation is acquired before mutation with a five-minute lease.
- Validation and DNS/favicon work occur before the functional transaction.
- Item, sharekey, TOTP, tag, custom-field, audit/revision, cache/health/task, queued WebSocket and
  idempotency-completion rows are committed together.
- Audit/revision handling is strict on these create/delete paths, and the returned revision pair is
  verified before commit.
- The item row is locked for delete and the optional revision is checked immediately before the
  mutation.
- External UDP syslog is emitted after commit, so it is never sent from a rolled-back transaction
  and is not repeated by an idempotent replay.
- A committed response survives a lost client connection. A crashed, uncommitted request rolls
  back and its expired lease can be taken over safely.

The remaining external-effect limitation is explicit: a process crash immediately after SQL
commit but before the post-commit UDP call may omit that syslog datagram. The durable TeamPass
audit row is still committed. After the 90-day cleanup window, an old key is intentionally treated
as a new operation.

## Documentation and tests

Updated OpenAPI, the basic API guide, API reference and revision architecture documentation, and
added a dedicated idempotency architecture document with replay/conflict/concurrency behavior and
`curl` examples.

Added `ApiItemIdempotencyTest` coverage for:

- key validation;
- canonical secret-keyed fingerprints;
- create/delete intent conflicts;
- replay, conflict, active-processing and stale-lease decisions;
- strict unsigned delete revision parsing;
- secret-free and correctly scoped schema contracts;
- transaction ordering and post-commit syslog placement;
- active-lease-safe cleanup;
- OpenAPI create/delete contracts.

Local validation completed:

- `app/api/openapi.json` parses successfully as OpenAPI 3.1.0.
- The create operation exposes one idempotency parameter and `409` response.
- The delete operation exposes idempotency, `id`, `revision`, deletion revision metadata and
  `409` response.
- Fresh-install and upgrade definitions contain the same idempotency column, table identity,
  lease/expiry fields and unique indexes; the installer action is wired to its table method.
- Source ordering guards confirm reservation completion precedes commit, and delete locking and
  revision validation precede mutation.
- No merge-conflict markers, `var_dump()` or `console.log()` were introduced in the changed
  source files.

Not executed in this local snapshot:

- PHP syntax checks, PHPUnit and PHPStan: no PHP executable is installed or available in `PATH`;
- WSL fallback: WSL is not installed;
- fresh-install/upgrade and role/folder runtime matrices: no disposable TeamPass database is
  available in this workspace.

Commands to run in the project CI/runtime environment:

```bash
php -l app/api/Model/ApiIdempotencyModel.php
php -l app/api/Controller/Api/ItemController.php
php -l app/api/Model/ItemModel.php
php -l app/sources/main.functions.php
php -l public/install/upgrade_run_3.2.2.php
php app/vendor/phpunit/phpunit/phpunit tests/Unit/Api/ApiItemIdempotencyTest.php
php app/vendor/phpunit/phpunit/phpunit
php app/vendor/bin/phpstan analyse --memory-limit=2G
```

guerricv:
CI follow-up after the implementation push:

- Pushed `ef9ad23b2` to unify `ItemModel::addItem()` exception handling. Validation errors still return `422`, while any exception raised after the transaction starts now follows the same rollback path.
- Tightened `ApiItemIdempotencyTest` so it checks only the internal `error_log()` payload for exception-message leakage, without rejecting the intentional validation message returned to API clients.

All checks now pass on `ef9ad23b233224f68558218fa6e3ddd0be2cc89c`: PHPUnit on PHP 8.2 and 8.3, PHPStan level 4, CodeQL, table-prefix validation, and the production-autoloader check.

nilsteampassnet:
Multiple contexts.
PR has been splited in several ones.

PR #5349, opened 2026-08-26
Add secure folder search, filtering and reset controls

guerricv:
## Description

Extend the modernized user search with folder support.

- Return matching folders in a visually distinct section above item results.
- Display each folder with an icon, its title and its authorized tree path.
- Add folder titles to the default text-search fields.
- Add a folder filter that restricts results to the selected folder and its descendants.
- Add a compact reset button beside the search bar.
- Make the existing "Clear all" action restore the complete default search state.
- Persist and restore the selected folder through the existing DataTables state.
- Provide translations for the new search controls in all 25 shipped languages.

Folder results and filter options use the same authorization scope as item search:

- Only folders accessible to the current user are considered.
- Foreign personal folders and all their descendants are excluded.
- A client-provided folder can only narrow the authorized scope, never extend it.
- Folder paths only contain ancestors present in that authorized scope.

## Related issue

N/A

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Documentation
- [x] Translation
- [ ] Refactor / maintenance

## How has this been tested?

- Tested in real-world conditions in a lab environment.
- PHPStan level 4 on the complete configured scope: passed with no errors.
- Search unit tests: 61 tests and 369 assertions passed.
- All 25 shipped language files pass PHP syntax validation.
- A regression test verifies that the three new controls have non-empty, non-English-fallback translations in every shipped language.
- Table-prefix guard: passed for all 80 known table names.
- `git diff --check`: passed.
- The complete PHPUnit suite ran 1,532 tests and 41,480 assertions. It reports 9 failures and 1 warning in unrelated existing tests on this Windows environment (`HealthSystemConsistencyTest`, LAPR, password-strength, security-posture and user-field authorization tests). No search test failed.

## Checklist

- [ ] The test suite passes (`php _tools/phpunit.phar` — see testing note above)
- [x] No new `app/sources/*.queries.php` file was added
- [x] No `teampassclasses` package was changed
- [x] `app/vendor/composer/` remains in its production form

- [x] No schema change

## Screenshots

Not attached.


nilsteampassnet:
Thanks for this, the authorization work is genuinely solid. Binding both `folder.id` and `ancestor.id` to `%li_folder_scope` so a breadcrumb can never leak an unauthorized ancestor title is exactly right, the fail-closed (`1 = 0`) paths are consistent, and the break you added in `searchBuildWhere()` avoids the trap I was expecting to find (a folder-only text search returning the whole faceted item scope). The containment-based `getForeignPersonalFolderIds()` is a real hardening beyond the feature itself. The saved-state versioning (`tpFiltersVersion = 2`) is a thoughtful touch.

I ran the full gate locally on the branch merged onto develop @ `912b04a6b`:
* `php _tools/phpunit.phar` → OK — 1533 tests, 41750 assertions
* PHPStan level 4 → 0 errors (223 files)
* `check_table_prefix.php` → OK — 80 table names
* `git diff --check` → clean

The 9 failures you saw are specific to your Windows environment, the suite is fully green here, so no concern on that front.

Two functional defects need fixing before merge, plus a few smaller points.

### Blocking

#### 1. Folder titles are stored HTML-encoded, neither decoded for display nor for matching

`nested_tree.title` goes through the sanitizer's escape filter on create (`app/sources/folders.queries.php:357` and `:784` → `Elegant\Sanitizer\Filters\EscapeHTML` → `htmlspecialchars(strip_tags(...))`). A folder named `R&D` is stored as `R&amp;D`; `Client's data` is stored as `Client&#039;s data`.

The new code returns the raw stored value (`app/sources/search.queries.php:250` and `:328`) and renders it with `.text()`. Two consequences:

- **Matching is broken**: this is the one that matters, and it's new here. `folder.title LIKE %ss_folder_term0` compares the user's input against the encoded form, so typing `R&D` or `Client's` returns nothing. That's the core of the feature.
- **Display**: `R&amp;D` shows literally in the results list, in the path breadcrumb, and in the filter `<select>`.

Every other title-rendering path in the codebase decodes first. `app/sources/items.queries.php:7591` builds the exact same `' / '` path format and does:

```php
$pathParts[] = htmlspecialchars(stripslashes(htmlspecialchars_decode($foldersById[$currentId]->title, ENT_QUOTES)), ENT_QUOTES);
```

and app/sources/tree.php:452 does the same for the tree.

In fairness, the display half already exists on the item side of this page (the Folder column at search.queries.php:516 is fed from cache.folder, which concatenates still-encoded titles in main.functions.php:1596). But the matching half is introduced by this PR, and it's what makes folder search silently fail on any folder with &, ', ", < or > in its name, which in practice covers a lot of real folder names.

Suggested fix: decode server-side before json_encode (the stripslashes(htmlspecialchars_decode(..., ENT_QUOTES)) form used elsewhere), and normalize the search terms the same way before they reach the LIKE.

#### 2. The personal root folder shows as a numeric user id
The title of a personal root folder is the user id. app/sources/tree.php:447-448 substitutes the login for display:

```php
// If personal Folder, convert id into user name
if ((int) $currentNode->title === (int) $inputData['userId'] && (int)$currentNode->nlevel === 1) {
    $currentNode->title =$inputData['userLogin'];
}
```

Folder search doesn't do this substitution, so a user sees 42 as a folder title in the results and 42 / Servers as the path of their personal subfolders, both in the result list and in the filter dropdown. No disclosure (only the caller's own personal tree is ever in scope), but it reads as a glitch. The same nlevel === 1 test applies here.

### Non-blocking

* **Unbounded folder option list** (`search.queries.php:233-245`): The `filter_options` query returns every accessible folder with a self-join and `GROUP_CONCAT`, with no `LIMIT`, while the tags query right below it, in the same block, is capped at `LIMIT 300`. On an instance with a few thousand folders that's a heavy query plus a very large `<select>` on every page load. Worth matching the neighbouring guardrail.
* **`group_concat_max_len` defaults to 1024 bytes**: A deep tree with long titles will have its path silently truncated, with no indication.
* **Two `NestedTree` rebuilds per request**: `getForeignPersonalFolderIds()` (`search.queries.php:202`) now runs on every draw of an endpoint driven by typing (300 ms debounce), and `getOwnPersonalFolderIds()` runs a second full rebuild at `:300` whenever `scope_perso` is set. The hardening is worth it, but a per-request memo would avoid paying for it on each keystroke.
* **Missing translation key**: `search_folder_results_more` is missing from 23 of the 25 language files, only `english.php` and `french.php` have it. `Language::getShipped()` falls back to English so nothing breaks, but your regression test only covers the other three keys, so this looks like an oversight rather than a decision.
* **Ambiguous label**: "Search in → Folder" is ambiguous. `searchSortColumnMap()` already exposes `6 => 'c.folder'`, but `searchTextFieldMap()` deliberately doesn't map folder. So ticking "Folder" does not find items whose folder name matches — only the folders themselves. It's documented, but the checkbox carries the same label as the results-table column, so some users will read it the other way.
* **Hardcoded limits**: `LIMIT 21` / `array_slice(..., 20)` are hard-coded at `:320` and `:326`, with no constant or setting.
* **Unconditional heading**: The `<h4>Items</h4>` heading renders unconditionally, including before any search has run while `#search-empty-hint` is showing.

---

Nothing here changes the design, the model and the test coverage are sound, and the two blocking items are localised (the two `'title' => (string) $row['title']` mappings at `search.queries.php:250` / `:328`, the term normalization feeding the `LIKE`, and the personal-root label).

Happy to re-review once those are in.

guerricv:
Hi Nils,

Thank you very much for the thorough review. Your observations were spot on and helped us cover cases that our original validation had missed.

We have now addressed all the points you raised:

- Folder search terms are normalized with the same HTML-encoding rule used when titles are stored. Titles and breadcrumbs are decoded server-side before safe text rendering, including the cached Folder column.
- Personal-root titles now display the current user's login instead of the numeric user id in results, breadcrumbs, and the folder filter.
- The folder filter is now a remote, ACL-bound Select2 with pagination (50 folders per page); restoring saved state resolves only the selected folder.
- Breadcrumbs are assembled in PHP from individually ACL-bound ancestor rows, removing the `GROUP_CONCAT` truncation risk.
- The repeated `NestedTree` work has been removed: subtree and personal-folder containment use SQL, and the existing personal-folder ids from the session are reused where possible.
- `search_folder_results_more` is now present in all 25 shipped language files.
- The Folder search label is now pluralized, the 20/21 limits are named constants, and the Items heading is hidden until a search is active.
- Regression coverage was extended for encoded titles (`&`, apostrophes, and quotes), personal-root labeling, ACL-safe paths, paginated folder options, translations, and the conditional heading. We also added a fail-closed guard for terms emptied by the folder-title sanitizer.

All CI checks completed so far are green; Scrutinizer is still running.

Thanks again for the precise and constructive feedback. It materially improved the robustness and scalability of the implementation. The PR is ready for another look whenever you have time.

nilsteampassnet:
Thanks @guerricv, the rework is solid.
Both blocking points are properly addressed: searchEncodeFolderTerm() fixes the matching half, not just the display half, and applying searchFolderDisplayTitle() to ancestors as well as the folder itself is the right call. Dropping GROUP_CONCAT for a row-set assembled in PHP, the paginated Select2, and removing both NestedTree rebuilds all go beyond what I asked for. Test suite (1538 tests) and PHPStan level 4 are green here.

One new blocker slipped in, in app/sources/search.queries.php:403-410:

`WHERE folder.id = %i AND folder.id IN %li_folder_scope`

That mixes a numbered placeholder (%i) with a named one (%li_folder_scope), which MeekroDB rejects at parse time, db.class.php:729, "You can't mix named and numbered args!". There's no try/catch around it, so filter_options returns a fatal instead of JSON whenever a folder filter is restored from the saved DataTables state (selected_folder > 0): the tags and custom-field dropdowns stay empty and the saved folder chip never resolves. The source-grep tests can't see it.

You already have the correct form 50 lines above in folder_options, so:

```
WHERE folder.id = %i_selected_folder_id AND folder.id IN %li_folder_scope
LIMIT 1',
['selected_folder_id' => $selectedFolderId, 'folder_scope' => $folderScope]
```

With that in, this is good to merge.

guerricv:
Hi Nils,

Thanks for catching this — we confirmed the mixed MeekroDB placeholder issue on the saved-folder restoration path.

We have now:

- merged the latest `develop` (`29c854a7f`) into the PR branch and preserved both the new upstream Search/Palette coverage and our folder-search tests;
- changed the selected-folder lookup to use `%i_selected_folder_id` and `%li_folder_scope` with one associative parameter array, so named and numbered arguments are no longer mixed;
- added a focused regression test that checks the corrected predicate is wired into the handler and passes that same predicate through `MeekroDB::parse()`. This exercises the parser rule that the previous source-shape tests missed.

The fix is pushed in `b1ba091aa`. GitHub now reports the PR as mergeable, and CI is restarting on the updated branch.

Thanks again for the precise review and for reproducing the failing saved-state path.

PR #5352, opened 2026-08-27
Unify file integrity and permission diagnostics in System Health

guerricv:
## Description

This PR replaces the legacy “unknown files” counter and web deletion workflow with one reliable, read-only diagnostic shared by the admin dashboard, System Health, background tasks and SSH tooling.

The implementation:

- parses the release `app/files_reference.txt` manifest defensively, supports MD5/SHA-256 entries and reports malformed, duplicate or unsafe paths;
- scans the current 3.2 application layout and separates modified, missing, unknown, legacy-layout, Composer development and scan-warning findings;
- excludes instance-owned runtime data and secrets from hash comparison while still detecting executable files or symbolic links planted in writable public avatar storage;
- applies one explicit repository-artifact policy to integrity and permission scans, making `.claude/`, `.github/`, documentation, tests and root development tooling neutral without broadly excluding hidden files;
- applies that same canonical policy in the release checksum generator so future manifests contain only runtime and deployment files;
- accepts an intentionally removed `public/install/` directory but fully verifies it whenever it is present;
- persists one atomic, version-aware report under `storage/logs/`, exposes stale/running/error states and paginates large result sets;
- runs asynchronously through the existing background-task system, with enqueue and scan locks preventing duplicate work;
- removes the web-side delete/ignore actions and provides reviewable, copyable `sudo` SSH cleanup commands instead;
- replaces the production cleanup suggestion based on a global `composer install` with a bundled offline CLI that reuses the deterministic `composer.lock` cleanup logic already used by install/upgrade;
- audits effective POSIX access and modes for every real TeamPass file and directory, distinguishing protected code, writable runtime paths, secrets and legacy instance data;
- detects Debian/Ubuntu and RHEL/Rocky/AlmaLinux/CentOS explicitly, maps the effective PHP/web identity, and generates hardened permission remediation only for those supported families (including SELinux runtime contexts on RHEL-family systems);
- uses the configured TeamPass regional date and time formats in the System Health page title;
- adds English/French UI strings, operator documentation and focused unit/consistency tests.

The permission audit follows the normal-runtime hardened posture. Upgrade-only paths such as `app/config/` remain read-only after remediation and must be unlocked temporarily before running the web upgrader.

`app/files_reference.txt` is intentionally not regenerated here: it remains a release artifact and should be refreshed by the release workflow once the final release tree is known.

## Related issue

N/A

## Type of change

- [x] Documentation
- [x] Translation
- [x] Refactor / maintenance

## How has this been tested?

- Added unit coverage for manifest parsing, modified/missing/unknown classification, legacy and development files, neutral repository artifacts, release-generator policy consistency, sensitive hidden files, writable avatar safeguards, installer removal, symbolic links, report persistence/pagination and safe SSH cleanup plans.
- Added permission-specific unit coverage for Ubuntu 26.04, Rocky/RHEL and unsupported distributions, non-Linux handling, recursive file/directory inspection, repository-artifact exclusions, protected sensitive dotfiles, world-writable detection, `sudo` remediation and RHEL SELinux commands.
- Added System Health consistency assertions for the shared report, admin-only endpoints, regional timestamp format, permission category, copy buttons and translated UI keys.
- Parsed the rendered Health JavaScript with V8 after replacing PHP interpolation blocks: syntax passed.
- Validated the release checksum generator with `bash -n`: syntax passed.
- Checked the Health markup structurally: 193 unique IDs, no duplicate IDs, and balanced `div`, `section`, table, list, button and select tags.
- Checked every referenced `TP_HEALTH_L10N` property: 187 references, no missing declaration.
- Performed lexical delimiter/string/comment checks on the changed PHP sources and tests: passed.
- Confirmed no `var_dump()` or `console.log()` was introduced and the obsolete web deletion actions remain absent.

## Checklist

- [ ] `app/vendor/composer/` is in its production form (`git checkout -- app/vendor/composer/`)

No new query handler or `teampassclasses` change is included, so the two corresponding checked items are not applicable rather than requiring duplicate files.

- [x] No schema change

The Health permission rules align with the paths already required by install/upgrade and extend them to the complete runtime tree. The generated remediation applies the hardened post-install posture; administrators must restore temporary write access to upgrade-only paths before launching an upgrade.

## Screenshots

image


nilsteampassnet:
## Review — Unify file integrity and permission diagnostics in System Health

**Type:** Feature + refactor + hardening · **Target branch:** `develop` ✅ · **Files:** 22 (+3712 / −753)

### Summary

Replaces the legacy "unknown files" counter and its web deletion workflow with a single **read-only** diagnostic shared by the admin dashboard, System Health, background tasks and CLI. The core is extracted from `admin.queries.php` (−477 lines) into three DB-free modules (`file_integrity.functions.php`, `file_permissions.functions.php`, `file_scope.functions.php`), run as a background task, with a persisted, paginated JSON report. Destructive web actions are gone, replaced by copyable `sudo` plans. Adds a full POSIX audit with Debian/RHEL-targeted remediation (including SELinux contexts).

### Overall verdict: ⚠️ Changes requested

Architecture, security and formal quality are excellent. **One sizing defect in the permission scan** should be fixed before merge.

---

### 🔴 Blocking

**1. The permission scan emits one finding per file → a 3.9 MB report decoded on every request**

Measured by actually running the scanner on a clean checkout:

```
Permission entries checked: 14205
Permission findings:        14207     ← of which 14181 are "protected_writable"
Report JSON:                3.9 MB
```

The trigger condition is `$writable && !$worldWritable` on the `protected` classification (`file_permissions.functions.php:438`). It is true **for every file** as soon as the tree is owned by the web account — an extremely common setup, and one TeamPass's own web upgrader requires (write access to `app/config/`, `csrfp.config.php`).

Measured per-request cost:

```
tpFileIntegrityLoadReport() + tpFileIntegritySummary()
  → 63 ms, 17.9 MB peak memory, 14207 entries held in RAM
    … just to render a summary card
```

`tpFileIntegrityLoadReport()` performs a full `json_decode` followed by `array_replace_recursive` over the whole report, and it is called:

- by `teampassGetSystemHealthOverview()` → **admin dashboard** (`admin.queries.php:3394`)
- by the Health page (`utilities.queries.php:989`)
- by `health_get_file_integrity_status`, which is **polled every 2 s** while a scan runs (`utilities.health.js.php:940`)
- by `health_get_file_integrity_issues`, which loads 3.9 MB to return 100 rows

Beyond the cost, the signal is unusable: "14207 permission findings" guides no action.

*Possible directions:* aggregate `protected_writable` per top-level directory (or store a count plus N samples); and above all, split a small summary file from the detailed report so the dashboard never has to decode `issues`.

---

### 🟡 Worth addressing

**2. Inconsistent scope between the two scans.** `tpFileIntegrityExcludedPaths()` excludes `storage/`, `secrets/`, `files/`, `upload/`, `backups/`; `tpFilePermissionsScan()` only filters repository artifacts and therefore **descends into every attachment and backup**. On a large vault: long walk plus extra `runtime_*` noise.

**3. Remediation strips the executable bit at the root.** `sudo find <paths> -xdev -type f -exec chmod 0640 {} +` (`file_permissions.functions.php:657`) applies to every top-level directory outside `storage|secrets|files|upload|backups`. `docker/` and `scripts/` are among them, so `docker/docker-entrypoint.sh` and `scripts/release_check.sh` lose `+x`. A Docker deployment bind-mounted from a checkout would break on the next start.

**4. `app/files_reference.txt` is not regenerated.** Acknowledged in the description, but the effect on a scan run today is concrete: **57 "modified" and the PR's own 5 new PHP files reported as `unknown [CRITICAL]`**, overall status `danger`. Regeneration by the release workflow is a hard prerequisite for shipping, otherwise every healthy installation shows a red health card.

**5. SQL `""` literal.** `finished_at = ""` (`utilities.queries.php`) breaks under the `ANSI_QUOTES` sql_mode (double quotes become an identifier). `finished_at = 0` against a `varchar(50)` column also relies on MySQL's loose numeric coercion. Works today, fragile.

**6. Orphaned `misc.unknown_files` row.** No PHP reader or writer left, and no upgrade script removes it. Harmless dead data.

**7. Variable shadowing.** `$referencePath` holds the manifest path (`file_integrity.functions.php:927`) and is then reused as the loop variable (`:934`). No current consequence.

**8. Dead language keys** (`delete_unknown_files`, `unknown_files_should_be_deleted`) kept across ~30 language files.

---

### 🟢 Strong points

- **Exemplary DB-free extraction**: testable logic isolated and reused by the web layer, the worker, the CLI and the tests. The 10 functions removed from `admin.queries.php` are a **clean** removal — no orphaned reference remains (only a guard test asserting their absence).
- **Single-source scope policy**: `tpFileScopeIsRepositoryArtifact()` is consumed by both the scanner and `regenerate_checksums.sh`, which makes manifest/scanner drift structurally impossible.
- **Removal of the destructive web path** in favour of human-reviewed commands — good security trade-off.
- **Path traversal properly handled**: `tpFileIntegrityNormalizePath()` rejects absolute paths, `..` and null bytes; `tpFileIntegrityIsInsideRoot()` confines symlinks; `RecursiveDirectoryIterator` does not follow links.
- **Remediation guardrails**: `escapeshellarg()` throughout, `web_user`/`web_group`/`codeOwner` validated by regex, whole-directory moves refused when the directory also holds reference files.
- **Report robustness**: atomic write (`rename`), shared `flock` on read, `schema_version`, explicit `stale`/`running`/`error` states, dual locking (enqueue + scan).
- **Detection of executable files planted in `public/assets/avatars/`** — a real attack angle, properly covered.

---

### PHPStan level 4

**✅ 0 errors** — run against this branch (`phpstan.phar`, `--memory-limit=2G`, project config, exit 0). The unchecked box in the description can be ticked.

### Tests

**✅ OK (1538 tests, 42089 assertions)** via `_tools/phpunit.phar`. That box can be ticked too.
The `NativeModalUsageTest.php` change (−2) only drops the assertion covering the deleted code — coverage is not weakened.

### Security

- Handlers correctly gated: session `key` check **and** `userAccessPage('utilities.health')`, plus `user-read_only` on scan start. `utilities.health` appears only in the `'admin'` list in `PerformChecks` → **admin-only endpoints confirmed**.
- **XSS: clean.** Every interpolated value goes through `tpEscapeHtml()`; `cleanup_plan` / `permissions_plan` are injected via `.text()`. The new PHP markup only echoes `$lang->get()`.
- **SQLi: clean.** MeekroDB placeholders (`%s`), `prefixTable()` throughout.
- Information disclosure (absolute paths, web account, code owner) is acceptable since it is strictly admin-only.
- No `var_dump()` / `console.log()` introduced. The modules touch neither the DB nor crypto — the RSA/sharekeys model is untouched.

### Install / upgrade impact

- ✅ No schema change — matches the declaration.
- ✅ No new `*.queries.php` → no `public/sources/` shim required.
- ✅ `app/vendor/composer/` untouched (0 files) → the third unchecked box is not applicable.
- ✅ Perfect EN/FR parity: 64 identical keys on both sides.
- ✅ `TP_HEALTH_L10N`: 187 references, 191 declarations, **none missing**.
- ⚠️ Release prerequisite: regenerate `app/files_reference.txt` (point 4).

---

### Recommended action

Address blocking point 1 (permission scan volume and report loading cost), then approve. Points 2–4 are worth handling in the same pass; 5–8 can follow.

guerricv:
Thanks for the detailed review. The findings have been addressed in `0e8c0fa44`, after merging the current `develop` branch (`29c854a7f`) into the PR branch.

Main adjustments:
- split persistence into a bounded `file-integrity-summary.json` for Dashboard/Health polling and the detailed report loaded only on explicit issue/CLI requests;
- publish details first and validate the shared `scan_id` before serving them;
- aggregate repeated permission findings by reason/scope while keeping exact counters and at most five sample paths (a synthetic 14,200-finding case now produces one group and a ~1 KB permission payload);
- keep the regular Health scan shallow for attachment, backup, avatar and legacy data contents while still checking their directory nodes; added CLI `--deep-permissions` for exhaustive audits;
- preserve executable bits in protected code remediation with `chmod u=rwX,g=rX,o=`;
- replaced the new `finished_at` comparisons with string placeholders compatible with `ANSI_QUOTES`;
- removed the shadowed `$referencePath` variable and the two unused legacy language keys;
- updated the UI and documentation for grouped findings and scan scope.

The orphaned `misc` rows are intentionally left for a separate idempotent data-cleanup migration. `app/files_reference.txt` is also intentionally unchanged: it remains a release artifact regenerated when a release is compiled.

Validation completed locally on PHP 8.2 and 8.3, with PHPStan level 4 clean. GitHub PHP 8.2/8.3 tests, CodeQL, JS analysis, table-prefix and production-autoloader guards are green; Scrutinizer is still running at the time of this comment.

guerricv:
Follow-up after the final quality pass:

- 52574917b now handles file-operation return values explicitly and keeps decoded payload types unambiguous.
- 72b2afdc replaces the ambiguous RecursiveCallbackFilterIterator callbacks with an explicit non-symlink directory traversal, while preserving repository exclusions and standard/deep permission-scan semantics.
- The final Scrutinizer comparison is down from 10 new findings to 2. All 8 findings in the file-integrity/permission module are gone; the 2 remaining findings are taint-flow variants on app/sources/main.functions.php:9289, a line authored in 2024 and outside this PR diff.
- GitHub CI is green for PHP 8.2, PHP 8.3, PHPStan level 4, CodeQL, JS/TS, table-prefix and production-autoloader checks.
- app/files_reference.txt remains intentionally unchanged; it stays a release-generation artifact.

PR #5354, opened 2026-08-29
Improve LAPR endpoint monitoring and rotation resilience

guerricv:
## Description

This PR improves the operational resilience, visibility, and administrator controls of Linux Account Password Rotation (LAPR).

It addresses two initial operational gaps:

1. Enrolled endpoint operating-system and privilege information previously remained unchanged after enrollment.
2. A scheduled rotation could fail when an endpoint was temporarily unavailable without providing a complete retry and suspension workflow.

The implementation also adds endpoint-level rotation pause/resume controls and integrates the resulting states into System Health and operational statistics.

### Enrolled endpoint checks

- Adds a manual **Check server** action to the enrolled-endpoints list.
- Runs availability, host-key, OS metadata, and effective `chpasswd` privilege checks entirely in the background worker.
- Refreshes the stored OS name, kernel metadata, root/sudo capability, last check, next check, and secret-free error state.
- Displays OS-specific remediation commands when the SSH identity can connect but no longer has the required rotation privileges.
- Adds configurable periodic endpoint checks:
  - `lapr_endpoint_checks_enabled`, enabled by default;
  - `lapr_endpoint_check_interval_minutes`, defaulting to 1,440 minutes.
- Uses the normal check interval after success and the existing transient retry delay for connectivity failures.
- Serializes checks, discovery, and rotations for the same endpoint through the background-task resource key.

### Rotation retry and failure handling

- Automatically retries only transient connectivity errors: timeout, connection refused, unreachable host, or lost connection.
- Uses the existing `lapr_max_retries` and `lapr_retry_delay_minutes` settings.
- Suspends the managed account after the retry limit is exceeded and requires an authorized operator to reset and resume it.
- Treats authentication, host-key, privilege, configuration, password-change, and synchronization errors as actionable failures instead of retrying indefinitely.
- Adds a customizable LAPR rotation-failure email template with endpoint, account, trigger, retry, and failure-state metadata. No password or SSH secret is included.
- Prevents automatic rotation when the latest endpoint check reports that rotation privileges are no longer valid.

### Endpoint-level pause and resume

- Adds a **Pause rotations** action with a fixed, secret-free reason list.
- Preserves managed-account states, retry counters, and rotation due dates while an endpoint is paused.
- Cancels pending rotations neutrally and prevents a worker already connecting from changing the remote password by re-reading the endpoint state immediately before `changePassword()`.
- Keeps non-destructive availability, OS, and privilege checks active at the normal cadence during a deliberate pause.
- Adds a **Resume rotations** action that keeps the endpoint paused until a background SSH and privilege check succeeds.
- Leaves existing due dates unchanged so overdue accounts become eligible on the next scheduler pass after a successful resume.
- Allows an explicitly confirmed manual break-glass rotation on a paused endpoint without resuming automatic rotations.
- Records pause, resume, and manual pause-override decisions in the LAPR audit log.
- Uses status-qualified updates so an in-flight endpoint check cannot overwrite a concurrent pause.

### System Health and statistics

- Reports active, paused, unreachable, and error endpoints separately.
- Adds a dedicated `endpoint_paused` managed-account state instead of treating an intentional pause as a generic error.
- Reports transient failures with a future retry as `retrying`.
- Avoids duplicating one endpoint outage as an error on every attached account.
- Adds endpoint-check health metrics: due, overdue, stale, pending, running, failed worker tasks, oldest pending task, and next scheduled check.
- Applies the global background-handler health status to endpoint checks even when automatic rotations are disabled.
- Extends Statistics with the endpoint status split and the endpoint-paused account state.
- Clarifies that rotation metrics count attempts, including retries, and deliberately avoids presenting sampled endpoint checks as an uptime percentage.

### UI, documentation, and translations

- Adds the new endpoint actions, confirmation dialogs, result panels, status labels, and remediation guidance to the LAPR pages.
- Makes manual rotation from both the managed-account list and item detail view aware of paused endpoints.
- Adds English and French strings for all new controls, states, warnings, health metrics, and statistics labels.
- Updates the LAPR user documentation and architecture notes.
- Seeds the two endpoint-check settings in the fresh-install and 3.2.2 upgrade paths.
- Extends the audit action-type comments for fresh installations; no table or column is added or altered.

### Safety properties

- SSH operations remain background-only; no SSH connection is opened by an HTTP request or by loading System Health.
- Endpoint ACLs and task ownership are re-evaluated when manual checks are polled.
- Pending tasks are de-duplicated using their indexed `item_id` relationship.
- Pauses, module disablement, account removal, and endpoint removal are checked again immediately before the remote password mutation.
- All new task results, audit events, and email tokens contain operational metadata only and never contain a password or SSH credential.

## Related issue

No linked issue.

## Type of change

- [x] Documentation
- [x] Translation
- [ ] Refactor / maintenance

## How has this been tested?

Manual end-to-end validation was completed by the contributor on the target TeamPass installation. The delivered endpoint check, OS refresh, privilege validation/remediation, pause/resume, manual break-glass rotation, retry, Health System, and Statistics flows were reported as functional.

Additional static validation performed before the commit:

- `git diff --check`: passed.
- Parsed all 25 changed PHP files with the PHP tree-sitter grammar: no syntax errors detected.
- Parsed all 5 changed JavaScript/PHP page scripts after replacing server-side PHP expressions: no JavaScript syntax errors detected.
- Verified that all 57 newly added language keys exist exactly once in both English and French.
- Checked the added code for hard-coded TeamPass table prefixes, HTTP-thread SSH calls, and obvious secret logging patterns: none found.
- Added unit and safety-wiring coverage for transient failure classification, endpoint check cadence, pause race protection, health classifications, cron overlay behavior, scheduler wiring, and operator notifications.

PHPUnit and PHPStan were not run locally because the available development environment did not provide PHP, Composer, or Docker. They remain intentionally unchecked below and should be confirmed by CI.

The exact manual role matrix was not recorded. LAPR operational actions retain the existing non-admin `can_manage_lapr` and item/folder ACL requirements. Personal-folder modes were not re-tested because LAPR-managed items and endpoint credentials are required to be non-personal by design.

## Checklist

- [x] New `app/sources/*.queries.php` files have a matching `public/sources/` proxy shim — not applicable; only existing handlers were changed
- [x] Changes to `teampassclasses` were applied to **both** copies (`app/includes/libraries/` and `app/vendor/`) — not applicable; no dual-location package was changed
- [x] `app/vendor/composer/` is in its production form (`git checkout -- app/vendor/composer/`) — the directory was not changed

- [x] No schema change

The fresh-install and 3.2.2 upgrade paths seed `lapr_endpoint_checks_enabled=1` and `lapr_endpoint_check_interval_minutes=1440`. Existing LAPR tables and enum values are reused. The installer and upgrade scripts were reviewed statically but were not executed locally.

## Screenshots

No screenshots are included in this draft. Recommended screenshots before submission:

- Enrolled endpoint list with **Check**, **Pause/Resume**, and **Delete** actions.
- Manual endpoint-check result showing refreshed OS information.
- Privilege-check failure with remediation commands.
- Endpoint pause confirmation and paused status.
- LAPR section in System Health.
- LAPR endpoint/account state charts in Statistics.


guerricv:
PHPStan level 4 findings have been addressed in commit `a2414b5e6`:

- simplified the paused-endpoint override flag after its confirmation guard;
- removed redundant `is_array()` checks for values guaranteed by the documented `testAndCollect()` return shape.

These are static-analysis cleanups only; the intended LAPR behavior remains unchanged.

nilsteampassnet:
Thanks for this, it is a well-structured contribution. The race handling around the pause (status-qualified writes plus the `$freshEndpointStatus` re-read immediately before `changePassword()`) is exactly right, the SSH-stays-in-the-worker guarantee is preserved on both new entry points, and separating enrollment tests from enrolled-endpoint checks by `item_id` is a clean way to keep the queue metrics unambiguous.

I ran the two checks you left unticked, on your branch:

* `php app/vendor/bin/phpstan analyse --memory-limit=2G` → [OK] No errors
* `php _tools/phpunit.phar` → OK (1525 tests, 41784 assertions)
* `php .github/scripts/check_table_prefix.php` → OK

I also verified that the 57 new language keys exist in both `english.php` and `french.php`, that every JS-side alias resolves, and that `next_check_at` / `last_error` / the `disabled` enum value were already in the DDL, so "no schema change" is accurate.

One blocking point and a few smaller ones below.

---

### Blocking

#### 1. An account-level `chpasswd` failure disables the whole endpoint

`app/scripts/traits/LAPRRotationTrait.php:308`:

```php
$change =$service->changePassword($username,$newPassword, $useSudo);$service->disconnect();
if ($change['success'] !== true) {$code = isset($change['error_code']) ? (string)$change['error_code'] : LAPRSshService::ERR_CHPASSWD_FAILED;
    $this->laprUpdateEndpointCheckState((int)$account['endpoint_id'], $actorId,$code);

```

`changePassword()` can only return `ERR_NOT_CONNECTED`, `ERR_INVALID_USERNAME` or `ERR_CHPASSWD_FAILED` (`LAPRSshService.php:327-348`). The last two are account-scoped — a PAM policy rejecting the new password, a locked account, a stale `username_cache` — but `laprEndpointStatusForCheckFailure()` maps them to `status = 'error'` on the shared endpoint.

Combined with the new scheduler gate (`background_tasks___handler.php:558`, which selects only `e.status = 'active'` or unreachable), one account whose password is refused by PAM silently stops automatic rotation for every other account on that host, and turns the endpoint red in System Health. Recovery waits for the next successful check, up to `lapr_endpoint_check_interval_minutes`, i.e. 24 h by default.

The three other call sites (`:202` credential unreadable, `:223` hostname not allowed, `:253` connect failure) are genuinely endpoint-scoped and correct; only this one is not. Restricting it to transport-level codes should be enough:

```php
if (laprIsTransientConnectionError($code) === true) {
    $this->laprUpdateEndpointCheckState((int)$account['endpoint_id'], $actorId,$code);
}

```

---

### Worth addressing

* **`ERR_UNKNOWN` is no longer retried, and it is a catch-all**: `laprIsTransientConnectionError()` lists exactly the four real constants, which is right. But `classifyConnectError()` (`LAPRSshService.php:405-422`) only matches `timed out|timeout|refused|no route|unreachable|getaddrinfo|unable to connect`; anything else becomes `ERR_UNKNOWN`. A mid-session socket drop ("Connection closed by server", "reset by peer", "Error reading from socket") therefore lands in `ERR_UNKNOWN` and now sends the account straight to `status = 'error'`, requiring a manual Reset & resume, where it used to be retried up to `lapr_max_retries`. Either add `ERR_UNKNOWN` to the transient list, or broaden `classifyConnectError()` so those messages map to `ERR_NOT_CONNECTED`.
* **Periodic checks are enabled by default on upgrade**: `public/install/upgrade_run_3.2.2.php:315` seeds `lapr_endpoint_checks_enabled = 1`. Existing installations will start opening outbound SSH connections to every enrolled endpoint every 24h without the administrator opting in, the opposite of `lapr_scheduler_enabled`, which is seeded at `0`. The checks are non-destructive, so this is defensible, but it is a behaviour change on upgrade. Either default it to `0` on the upgrade path and `1` on fresh install, or flag it explicitly so it can go in the release notes.
* **No throttling on the failure alert email**: `laprSendRotationFailureAlert()` fires on every attempt. With `lapr_max_retries = 3` a single outage produces 4 emails per account; a host with 20 managed accounts going down produces 80. Consider sending only on the `suspended` and `action_required` states, or deduplicating per endpoint per outage.
* **Misleading message for a paused account**: `app/sources/lapr_accounts.queries.php:829`, the guard moved from `a.status != 'deleted'` to `a.status IN ('active','error')`. The hardening is correct (it now matches the `canRotate` rule in the UI), but a paused account now gets `lapr_account_not_found`, which points the operator in the wrong direction. A dedicated "reset the account first" message would be clearer.
* **One extra query per rotation candidate in the drain loop**: `background_tasks___handler.php:1467-1476` resolves `endpoint_id` from `lapr_accounts` for each `lapr_rotation` candidate on every drain pass, up to 100 per tick. Correct and cheap, but it could be folded into the candidate `SELECT` that already joins the `accounts` table.
* `laprPauseEndpoint()` swallows the exception in its `catch (Throwable $e)` with no log line.
* `laprUpdateEndpointCheckState()` writes `last_check_at` on every rotation, so the new stale metric in `laprMonitoringEndpointChecks()` under-reports for frequently rotated endpoints — they never look stale even if no check ever ran.
* The scheduler lease uses `lapr_retry_delay_minutes`; if the check interval is configured below the retry delay, a crashed worker delays the next attempt by the longer of the two.



---

Everything else looks good to me, the authorization chain matches the existing endpoint-management model, no secret reaches `last_error`, the audit details, or the email tokens, and the remote-controlled values in the new JS are all constrained (`os_family` against a fixed list, `ssh_username` against a regex, every sink through `laprHtml()` / `DOMPurify` / `tpEscapeHtml`).

Once fixed I am happy to merge.

guerricv:
@nilsteampassnet Thank you for the detailed review. All the blocking and “worth addressing” points have now been handled in commit `a313e5b91`.

### Blocking fix

- A post-connect `chpasswd` rejection is now strictly account-scoped. `ERR_INVALID_USERNAME` and `ERR_CHPASSWD_FAILED` no longer move the shared endpoint to `error` or block rotations for its other accounts.
- Only a classified transport loss during the password-change command updates the endpoint availability and enters the scheduler retry flow.

### Additional review fixes

- Broadened SSH transport classification for DNS failures, closed/reset connections, broken pipes and socket read/write failures. These map to the existing transient error taxonomy instead of falling through to terminal `ERR_UNKNOWN`.
- Normalized phpseclib `exec()` failures and retained the first transport failure during multi-command OS/capability or account-discovery operations, so partial results are never reported as successful checks.
- Periodic endpoint checks now default to disabled on upgrade (`0`) while remaining enabled for fresh installs (`1`), avoiding unsolicited outbound SSH traffic after an upgrade.
- Rotation alerts are throttled: terminal/action-required failures are sent immediately; transient outages notify on the first retry and again only on final suspension.
- A paused managed account now returns a dedicated operator message directing the user to **Reset & resume**, instead of the misleading “account not found”.
- Rotation resource resolution no longer performs one query per candidate. New tasks carry a server-computed `endpoint_id`; legacy queued tasks are resolved in one batch.
- Exceptions from the endpoint-pause transaction are now logged server-side while the UI keeps the generic safe error response.
- Rotation observations no longer update `last_check_at` or postpone `next_check_at`; a transient failure may only bring the next full endpoint check forward.
- The endpoint-check enqueue lease now uses the shorter of the normal check interval and retry delay, so a crashed worker cannot delay the configured cadence.

### Localization

The complete PR language scope was also aligned:

- 58 PR-related keys covered (54 initial additions, the paused-account message, and 3 updated statistics labels);
- all 25 language catalogs contain each key exactly once;
- every non-English catalog now has a contextual translation rather than an English fallback;
- placeholders such as `#count#`, `#endpoint#`, `#retry_at#` and the email HTML structure are verified against the English source.

### Validation performed

- Full PHPStan level 4 after the behavioral fixes: **[OK] No errors**
- PHP syntax checks for all changed PHP files and all 25 language catalogs: **OK**
- `LaprFunctionsTest`: **72 tests, 154 assertions — OK**
- Affected LAPR safety/wiring tests: **5 tests, 115 assertions — OK**
- LAPR localization checks: **4 tests, 16,664 assertions — OK**
- Table-prefix guard: **OK — no hard-coded table name found (80 table names checked)**
- UTF-8 validation for all 25 catalogs: **OK**
- `git diff --check`: **OK**

A complete PHPUnit run was also exercised on Windows. Its remaining failures were limited to the known Windows/environment conditions (CRLF-sensitive source assertions, the checked-out DataTables symlink, missing LDAP support and unavailable Unix `rm`); the affected LAPR tests are green.

nilsteampassnet:
@guerricv, all 8 points are fixed and verified on `a313e5b91`.

The blocking one is properly scoped now: only a classified transport loss during `changePassword()` touches the endpoint, so a PAM rejection or a stale username no longer stops rotation for the other accounts on the host.

Nice extras: `getLastTransportFailure()` prevents a dropped channel from being reported as a successful check with partial OS metadata, and each fix has its own sentinel test.

On my side:

* `phpstan --memory-limit=2G` → [OK] No errors
* `_tools/phpunit.phar` → OK (1539 tests, 57768 assertions)
* `check_table_prefix.php` → OK
* no merge conflict with `develop`

Thanks for the thorough follow-up and for translating the 25 catalogues by hand, merging into `develop`.

PR #5358, opened 2026-09-03
Add idempotent item mutations and conditional delete

guerricv:
## Context

This is the second, deliberately isolated follow-up from #5343. The first part (API password history and durable revision timestamps) was integrated through #5348.

This PR adds the remaining "idempotent create/delete and conditional delete" scope for future bidirectional offline synchronization. It does not change mobile or browser-extension code.

## API contract

- `POST /item/create` accepts an optional `Idempotency-Key` containing 1-128 visible ASCII characters without spaces.
- The first keyed create keeps the normal `201`, `Location`, `newId`, `revision` and `revision_changed_at` response.
- Replaying the same user/key/canonical intent returns the stored response with `Idempotency-Replayed: true`, without repeating database writes or side effects.
- Reusing a key with a different intent returns `409`; an identical request with an active processing lease returns `409` and `Retry-After`.
- `DELETE /item/delete` accepts an optional unsigned `revision` precondition. A stale revision returns `409` before mutation; omitting it preserves the existing last-writer-wins behavior.
- Successful deletes return the exact deletion `revision` and `revision_changed_at` pair exposed by `GET /item/changes`.
- Delete supports the same optional idempotency header. Replaying a completed delete never performs the deletion again, including after a later restore.

Requests without `Idempotency-Key`, and deletes without `revision`, remain backward compatible.

## Persistence and security

- Adds the prefixed `api_idempotency` table, scoped by the unique tuple `(user_id, operation, key_hash)`.
- Adds nullable unique `items.api_idempotency_id` so a stale create reservation can recover an already committed item without repeating its side effects.
- Raw keys, request bodies, passwords, TOTP secrets and custom-field values are never persisted in idempotency metadata.
- Key hashes and canonical request fingerprints use HMAC-SHA-256 with a domain-separated key derived from the TeamPass instance secret.
- Create fingerprints cover every functional field, recursively sort associative keys and preserve list order.
- Completed records have a 90-day replay window. Maintenance clears expired links and records without deleting an active processing lease.

## Atomicity and crash behavior

- A five-minute processing lease is acquired before mutation.
- Validation and favicon/DNS work complete before the functional create transaction.
- Item, sharekey, TOTP, tag, custom-field, audit/revision, cache/health/task, queued WebSocket and idempotency-completion rows commit together.
- Audit/revision handling is strict for these transactional create/delete paths, and the returned revision pair is verified before commit.
- Delete locks the item row and evaluates the optional revision immediately before mutation.
- External UDP syslog emission occurs only after commit and is never repeated by a replay.
- A lost HTTP response can be replayed. An uncommitted crashed request rolls back, and its expired lease can be taken over.

The remaining external-effect limitation is explicit: a process crash immediately after SQL commit but before the UDP call can omit that syslog datagram. The durable TeamPass audit row is still committed.

## Install, upgrade and documentation

- Fresh-install schema and `public/install/upgrade_run_3.2.2.php` both include the new nullable item link, table and indexes.
- The upgrade remains idempotent.
- CORS, OpenAPI 3.1, the basic API guide, API reference and architecture documentation are updated.
- `ApiItemIdempotencyTest` covers key validation, canonical fingerprints, replay/conflict/lease decisions, delete revision parsing, schema security, transaction ordering, cleanup safety and OpenAPI contracts.

## Local validation

Completed successfully:

- verified that all 22 supplied PR5343 files match the original PR head;
- isolated the final three part-2 commits and confirmed them with `git range-diff`;
- parsed `app/api/openapi.json`;
- ran structural assertions over replay short-circuits, transaction ordering, row locking, revision checks, post-commit syslog and install/upgrade schemas;
- ran an equivalent of the table-prefix guard over all 81 known table names;
- verified the committed Composer autoloader remains production-only;
- ran `git diff --check` and checked for conflict markers and forbidden debug calls.

Not executed locally because this Windows workspace has no PHP CLI, Docker runtime, WSL installation or disposable TeamPass database:

- PHP syntax checks;
- PHPUnit;
- PHPStan;
- fresh-install/upgrade and runtime role/folder matrices.

The GitHub Quality and CodeQL workflows are expected to provide the PHP 8.2/8.3, PHPStan level 4 and security validation.

## Checklist

- [x] No `var_dump()` or `console.log()` introduced
- [x] No hard-coded TeamPass table prefix introduced
- [x] Fresh-install and upgrade paths are both updated
- [x] No `teampassclasses` dual-location class is changed
- [x] `app/vendor/composer/` remains in production form
- [x] PHPUnit passes on PHP 8.2 and 8.3 (CI)
- [x] PHPStan level 4 passes (CI)
- [ ] Database-backed install/upgrade scenarios are exercised

## Screenshots

Not applicable. This PR changes API, database-schema and backend behavior without changing the web UI.

### DIFF
--- a/.claude/docs/api-reference.md
+++ b/.claude/docs/api-reference.md
@@ -249,12 +249,13 @@ Create a new item.
 
 **Optional idempotency:** `Idempotency-Key: <opaque-key>` (1–128 visible ASCII characters,
 without spaces). Its identity is scoped to the authenticated user and `item.create`. The first
-accepted request returns the normal `201`; an identical replay within 90 days returns the exact
-stored result and `Idempotency-Replayed: true` without repeating any write or side effect. Reusing
-the key with a different functional payload returns `409`. A concurrent request still holding the
-key's processing lease also returns `409` with `Retry-After`. The fingerprint covers every
-functional create field, including credential, TOTP and custom-field values, but is a
-server-secret HMAC: neither those values, the request body nor the raw key is persisted.
+accepted request returns the normal `201`; an identical replay within the configured offline-sync
+window (`offline_sync_window_days`, 90 days by default, `0` for no limit) returns the exact stored
+result and `Idempotency-Replayed: true` without repeating any write or side effect. Reusing the key
+with a different functional payload returns `409`. A concurrent request still holding the key's
+processing lease also returns `409` with `Retry-After`. The fingerprint covers every functional
+create field, including credential, TOTP and custom-field values, but is a server-secret HMAC:
+neither those values, the request body nor the raw key is persisted.
 
 **Custom fields:** `fields` = array of `{ id, value }` (field id + value). Encrypt-before-INSERT for encrypted categories; creator sharekey created synchronously, other users via the `new_item` background task. Only fields tied to the folder are stored; empty values ignored. Requires `item_extra_fields`.
 
@@ -323,9 +324,10 @@ Soft-delete an item.
 **Params:** `id` (int), optional `revision` (unsigned 32-bit integer). When supplied, `revision`
 is an optimistic-concurrency precondition checked against the locked item row immediately before
 the mutation. A mismatch returns `409` with no delete, audit, revision, cache change or WebSocket
-event. Omitting it preserves the historical last-writer-wins behavior.
+event. Omitting it or sending an empty query value preserves the historical last-writer-wins
+behavior.
 
-**Optional idempotency:** `Idempotency-Key` follows the same syntax and 90-day replay window as
+**Optional idempotency:** `Idempotency-Key` follows the same syntax and configured replay window as
 create, scoped to the authenticated user and `item.delete`. Its fingerprint contains the item id
 and the expected revision (including omission). A replay returns the original successful result
 with `Idempotency-Replayed: true`; it never deletes again, so a later restoration is not undone.
--- a/.claude/docs/api-reference.md
+++ b/.claude/docs/api-reference.md
@@ -247,9 +247,18 @@ Create a new item.
 
 **Body:** `label`, `password`, `folder_id`, optional `description`, `login`, `email`, `url`, `tags`, `totp`, `fields`.
 
+**Optional idempotency:** `Idempotency-Key: <opaque-key>` (1–128 visible ASCII characters,
+without spaces). Its identity is scoped to the authenticated user and `item.create`. The first
+accepted request returns the normal `201`; an identical replay within 90 days returns the exact
+stored result and `Idempotency-Replayed: true` without repeating any write or side effect. Reusing
+the key with a different functional payload returns `409`. A concurrent request still holding the
+key's processing lease also returns `409` with `Retry-After`. The fingerprint covers every
+functional create field, including credential, TOTP and custom-field values, but is a
+server-secret HMAC: neither those values, the request body nor the raw key is persisted.
+
 **Custom fields:** `fields` = array of `{ id, value }` (field id + value). Encrypt-before-INSERT for encrypted categories; creator sharekey created synchronously, other users via the `new_item` background task. Only fields tied to the folder are stored; empty values ignored. Requires `item_extra_fields`.
 
-**Response 201:** `{ error: false, message, newId, revision, revision_changed_at }` + `Location: /api/v1/item/get?id=<newId>` (path-absolute reference). Validation failures → `422`; missing fields → `400`; folder not allowed / read-only → `403`.
+**Response 201:** `{ error: false, message, newId, revision, revision_changed_at }` + `Location: /api/v1/item/get?id=<newId>` (path-absolute reference). An idempotent replay preserves both. Validation failures → `422`; missing fields or invalid idempotency key → `400`; folder not allowed / read-only → `403`.
 
 **Permissions:** `allowed_to_create`. Blocked with 403 if folder is read-only for user.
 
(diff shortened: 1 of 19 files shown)


guerricv:
CI follow-up for `3edeb790f`:

- All GitHub Actions checks pass, including PHPUnit on PHP 8.2/8.3, PHPStan level 4, CodeQL, and the repository guards.
- Scrutinizer completed successfully and its tests passed. Its report lists 5 new issues and 22 updated code elements; these are non-blocking under the current check configuration.

The PR checklist has been updated for PHPUnit and PHPStan. The database-backed fresh-install/upgrade scenario remains intentionally unchecked because it was not exercised locally.

guerricv:
Scrutinizer follow-up for `cd4bbe3af` and `311989f9e`:

The findings attributable to this PR have been addressed without changing functional behavior:

- made the `file_get_contents()` `string|false` guard explicit before passing the secret to `hash_hmac()`;
- removed all dead `$transactionStarted = false` assignments immediately preceding early returns in `ItemModel::deleteItem()`;
- removed the dead `$transactionStarted = false` assignment immediately before returning from `pruneApiIdempotencyRecords()`.

The first rerun cleared the HMAC and pruning findings. Scrutinizer then moved the same `deleteItem()` issue identifier to the next equivalent early-return branch, so `311989f9e` removes all three remaining occurrences of that pattern.

The final Scrutinizer inspection passes and now reports only 2 new issues. They are the pre-existing security findings on the unchanged email body path in `main.functions.php`: former line-9293 identifiers `50930637`/`50930638`, recreated at line 9385 as `50933548`/`50933549` after this PR shifts the file. They are not caused by the idempotency implementation.

`git diff --check` passes locally. The final CI is fully green: PHPUnit on PHP 8.2 and 8.3, PHPStan level 4, CodeQL, all repository guards, and Scrutinizer.

nilsteampassnet:
Thanks for the careful isolation from #5343, this reviews well. Locally on develop: PHPStan level 4 clean, PHPUnit 1688 tests green, check_table_prefix.php picks up the new table on its own, openapi.json parses with every $ref resolving, and install/upgrade are complete and idempotent. No raw key, body, password, TOTP secret or field value is persisted, and the reserve/replay/conflict/lease state machine looks correct.

🔴 One blocker: storeTask() now runs inside the create transaction.
handlePostInsertTasks() ends with storeTask('new_item', …) (ItemModel.php:1156), which itself ends with triggerBackgroundHandler() (main.functions.php:8848, an exec() spawn), while the commit only happens at ItemModel.php:593. The handler is launched before the background_tasks rows are visible, finds nothing and exits, so the FUNC-1 sharekey fan-out waits for the next cron tick instead of starting immediately, and until it runs, other users get the documented 422 on a password update. The web path avoids this deliberately: items.queries.php commits at :585 and calls storeTask() at :782. Please move the call after DB::commit(), next to emitItemSyslog().

🟡 Points to address in the same pass:
1. Lock-order inversion: create takes api_idempotency then items; pruneApiIdempotencyRecords() takes them the other way round, deadlock possible, the prune swallows it but the create returns 500.
2. error_log() no longer records the exception message or file/line, so a 500 is undiagnosable. MeekroDBException::getMessage() is only $db->error (the SQL lives in getQuery()), so logging it leaks nothing. The unit test locks this choice and needs updating too.
3. refreshItemHealthAfterSave() is inside the transaction (ItemModel.php:573) and runs two unfiltered GROUP BY derived tables over all of log_items and sharekeys_items → it is a cache refresh and can go after the commit.
4. REPLAY_WINDOW_SECONDS is hard-coded at 90 days while the comment claims alignment with offline_sync_window_days, which is admin-configurable.
5. revision= (empty value) returns 400; many clients emit an empty parameter for an absent value → treating it as omitted would be friendlier.
6. architecture-api-idempotency.md is not referenced from CLAUDE.md, unlike every other architecture doc, so it is never loaded.
7. The tests are mostly strpos() assertions over source files: they break on any harmless refactor and cover none of reserve() / replay / lease takeover. Extracting the state machine into a DB-free *_logic.php module (as item_revisions_logic.php does) would let it be tested for real.
8. The items ALTER is done twice (column, then unique key)→ merging them halves the rebuild on large installations.

Two smaller ones, no action required: the duplicated authorization check in deleteItem() deserves a comment saying the re-check under row lock is deliberate, and Retry-After on a 409 is non-standard.

Nice side effect worth keeping: deleteItem now emits the item_deleted WebSocket event that was missing before.

guerricv:
Review follow-up implemented in 093a1a364 and dde8fe8dd:

- Blocker: the new_item task is now created and its detached handler launched only after the create transaction commits, when the item and replay result are visible.
- Lock order: idempotency cleanup now locks expired api_idempotency rows first, then clears item links, matching mutation order.
- Diagnostics: create/delete 500 logs again include the exception message, file and line while responses remain generic.
- Transaction scope: the item-health cache refresh now runs after commit.
- Replay lifetime: expiry follows offline_sync_window_days; 90 days remains the default and 0 means no expiry.
- Conditional delete: an empty revision query value is treated as omitted and documented in OpenAPI.
- Architecture: architecture-api-idempotency.md is now referenced from CLAUDE.md and updated for the post-commit behavior.
- Tests: the reserve/replay/conflict/processing/stale/takeover decisions and configurable expiry now live in a DB-free api_idempotency_logic.php module with behavioral PHPUnit coverage.
- Upgrade: the normal items schema path adds api_idempotency_id and its unique key in one ALTER, while retaining retry-safe partial-upgrade branches.

Also added the optional explanatory comment for the deliberate authorization re-check under the locked item row.

Validation on the latest commit dde8fe8dd:
- PHPUnit on PHP 8.2: success.
- PHPUnit on PHP 8.3: success.
- PHPStan level 4: success.
- CodeQL, table-prefix guard, and committed-autoloader guard: success.
- OpenAPI JSON parses locally and all internal references resolve.
- Scrutinizer: success (analysis completed, 25 updated code elements; tests passed).

The first CI pass caught a missing passwordKey argument introduced while moving the background task. Commit dde8fe8dd restores that creator share-key input; the complete second GitHub Actions run is green.

PR #5368, opened 2026-09-08
Fix monitoring log filters and clarify purge controls

guerricv:
## Description

The Items log search sent a selected column that the server ignored: searching `alice` in User also returned Bob's activity on an item called `alice portal`. Apply an explicit column allow-list, include the displayed name fields for User, translate the selector and preserve its selection with the saved table state. Remove Date, API and Personal from the search selector while keeping their display columns; old saved selections fall back to All columns. Exclude raw timestamps from global search and match action translations or codes through a parameterized action predicate, including manual and password-access events.

The footer's User dropdown is a deletion criterion. Clarify its purpose across the log tabs and fix the Errors, Copy, Admin and Failed logins purge paths that ignored it. Failed attempts are matched by the selected account's current login, including recognized API log entries. A purge restricted to one account retains attempts recorded under a different login before an account rename. Matching follows the database collation (normally utf8mb4_unicode_ci), so case variants such as alice/Alice are included. Use parameterized deletion scopes, reject invalid criteria, include the final calendar day and clear confirmation when criteria or tabs change. Keep administrators in the account selector and exclude the four internal accounts.

Knowledge Base logs now translate action codes and known details through the language catalogs, display the user's full name with their login, and search those displayed values. Missing accounts retain the stored login. Complete the six missing Arabic labels.

## Related issue

Related to #5363. The report does not identify the exact selector; the footer ambiguity and the independently confirmed defects are addressed together.

## Type of change

- [ ] Documentation
- [x] Translation
- [ ] Refactor / maintenance

## How has this been tested?

CI verified on commit `2f7e9cdfa5ae33189174ad267add695324398e4f`. GitHub Actions tested the PR merge with develop (`540fa721fb8b453673f9e219593ce90ad6472f7e`):

- [Quality workflow](https://github.com/nilsteampassnet/TeamPass/actions/runs/34578042696): the full suite passes on **PHP 8.2 and 8.3, with 1,803 tests and 60,176 assertions on each version**, with SQLite enabled. PHPStan level 4, all 45 login JavaScript tests, table-prefix and production-autoloader checks also pass.
- [CodeQL workflow](https://github.com/nilsteampassnet/TeamPass/actions/runs/34578042712): JavaScript/TypeScript and GitHub Actions analyses pass; the CodeQL result check also succeeds.
- [Scrutinizer](https://scrutinizer-ci.com/g/nilsteampassnet/TeamPass/inspections/06e370d8-43ce-4671-a2b3-79a0857e75c6): successful status on the same commit; its GitHub status reports seven updated code elements and passing tests.

Local validation also passed:

- 29 targeted tests and 292 assertions passed on each PHP version with SQLite. Without SQLite, 20 tests pass (245 assertions) and six SQLite methods are skipped. Database-free tests remain enabled in both environments. Coverage includes translated action searches, timestamp exclusion, column isolation, displayed names, safe output, purge isolation, login case variants and pre-rename retention, invalid criteria and daylight-saving transitions. SQLite fixtures run entirely in memory and model ASCII login case folding, not the complete MySQL Unicode collation. Production has no new database dependency.
- Helpers are tested directly, with source assertions for handler wiring; no test evaluates production-source slices. Six V8 scenarios with simulated PHP template values verify the five remaining Items options, saved-selection fallback, request/state agreement and duplicate-handler prevention.
- Full PHPStan level 4, PHP syntax on 8.2/8.3, table-prefix and whitespace checks.
- The full local Windows PHP 8.3 suite retained nine previously observed failures. The Linux CI for this revision passes the complete suite on both PHP versions.

The CI retains existing maintenance warnings about action deprecations and Composer lockfile freshness; all checks completed successfully.

Manual instance/browser validation remains pending: administrator/manager/standard/read-only roles, personal folders on/off, Knowledge Base display and actual web/API log rows. The in-app browser could not start, so no browser validation is claimed.

## Checklist

- [x] PHPStan level 4 passes
- [x] The full test suite passes on PHP 8.2 and 8.3
- [x] No new `var_dump()` or `console.log()` in production
- [x] Public query proxy requirement — no new query endpoint
- [x] Dual-location class requirement — no class changes
- [x] `app/vendor/composer/` remains in its production form
- [ ] Manual instance/browser validation

- [x] No schema change
- [ ] Schema change

No installer or upgrade migration is needed. Existing language catalogs supply column labels; the seven new strings are translated in all 25 language catalogs. The Arabic catalog now returns its array to the current language loader while preserving its legacy global variable.

Translation checks passed: all 175 new strings are present once, non-empty and returned by the actual language loader; all catalogs are valid UTF-8 and pass PHP 8.2/8.3 syntax checks. The 15 action/detail labels used by Knowledge Base logs are also present and returned correctly in all 25 catalogs, including the six Arabic additions.

## Screenshots

Pending browser verification.


nilsteampassnet:
Hi @guerricv,

Thanks for the PR and for digging into #5363. The purge rework is a clear improvement: the deletion scope is now parameterized and fails closed, the 4 tabs that ignored the selected user are fixed, the last calendar day is included, the Admin tab also purges `user_mngt`, and the two selectors no longer  render raw HTML. I merged it locally on top of the current `develop`: the merge is clean and PHPStan level 4 passes on the changed files.

A few changes are needed before merging.

## Required

### 1. `LogsFilteringTest` errors when `sqlite3` is not installed

The CI passes only because `quality.yml` now installs the extension. Without it, the full suite ends with 11 errors (`Class "SQLite3" not found`). This includes the tests that never touch a database, because `setUp()` opens SQLite for every test.

- Move the pure tests (`testColumnSelectionOnlyAcceptsTheUiAllowList`, `testInvalidOrUnresolvablePurgeScopeIsRejected`, `testDateRangeIncludesTheLastDayAndHandlesDst`) to a separate class without the SQLite `setUp()`, so they run everywhere.
- Mark the remaining SQLite class with `#[RequiresPhpExtension('sqlite3')]`, so its tests are skipped instead of failing.

## Please also address

### 2. Two Items search columns still cannot match what the table shows

- **Date** searches the raw Unix timestamp (`l.date LIKE %ss`). A date typed as displayed never matches, and in *All columns* a short number adds noise: searching the item ID `42` also returns every row whose timestamp contains `42`. Please remove `l.date` from the selector and from *All columns*.
- **Action** searches the stored code (`at_creation`), while the table shows its translation. You already solved this for the Knowledge Base logs, and the same approach works here: map the known item action codes to their translated labels, keep the codes whose label (or code) contains the term, and add `l.action IN %ls`. A small DB-free helper in `logs_filter_logic.php` would keep it testable.

### 3. Please don't `eval()` slices of production files in tests

`searchItems()` and `knowledgeBaseLogData()` cut the handlers' source on comment markers and `eval()` the result. No other test in the suite does this, and renaming one of those comments breaks the test with an unrelated error (`strpos()` returns `false` and the wrong slice is evaluated).

- Test the helpers directly (`getItemLogSearchColumns()`, `buildLogsPurgeFilter()`), and cover the wiring with a source assertion, as `UtilitiesLogsEncodingTest` already does (for example, assert that the handler calls `getItemLogSearchColumns(`).
- Move the KB row display and search logic (the `$reasonKeys` mapping, `user_display`, `action_display`, `reason_display`) into a small DB-free function, and test that function.
- Don't assert hard-coded French strings (`'Vu'`, `'Intitulé'`, `'Éléments associés'`). Compare with `$lang->get('at_shown')` and so on; otherwise the next POEditor rewording breaks the test.

### 4. License header

`app/sources/logs_filter_logic.php` needs the standard TeamPass header (GPL block, `@file`, `@author`, `@copyright`, `@license`, `@see`), like the other `*_logic.php` modules. `item_update_logic.php` can serve as a template.

### 5. `var` in the new JavaScript

The new code in `showItems()` uses `var` three times. The convention is `const`/`let`, and the file already uses them: `const columns`, `let searchColumn` (it is reassigned), `const $select`.

### 6. Internal accounts in the purge user list

Switching to `WHERE id > 0` rightly adds the administrators, which the Admin tab needs. It also lists the internal accounts (OTV, TP, SSH, API), which are not meant to be selected here. Please exclude them:

```php
$rows = DB::query(
    'SELECT id, login, name, lastname FROM ' . prefixTable('users') . '
    WHERE id > 0 AND id NOT IN %li ORDER BY login',
    [(int) OTV_USER_ID, (int) TP_USER_ID, (int) SSH_USER_ID, (int) API_USER_ID]
);
```

### 7. Failed logins purge: state the limit and test the shipped behaviour

- The purge matches the account's **current** login, so attempts recorded before a rename are kept. That is acceptable, but please state it in the PR description (it will go into the changelog) and in the comment next to the query.
- In production, `log_system` uses `utf8mb4_unicode_ci`, so `field_1 = 'alice'` also deletes `Alice`. The SQLite fixture compares case-sensitively, so the test does not describe what we ship. Please declare the fixture column as `field_1 TEXT COLLATE NOCASE` and add a case-variant row to the assertions.

Thanks again, the rest looks good to me.

guerricv:
Hi @nilsteampassnet, thanks for the detailed review. All seven points from [your review](https://github.com/nilsteampassnet/TeamPass/pull/5368#issuecomment-5630257187) are addressed in 2f7e9cdfa:

1. **SQLite dependency:** the three pure tests now live in `LogsFilterLogicTest`. `LogsFilteringTest` uses `#[RequiresPhpExtension('sqlite3')]`. Local checks on PHP 8.2 and 8.3 confirm that the database-free tests run without SQLite and the database tests skip cleanly; SQLite remains enabled in CI.
2. **Items search:** Date is removed from the selector and global search, while its display and sorting remain available. Saved selections for removed fields fall back to All columns. Action search matches translated labels or stored codes through a database-free helper and a parameterized `IN` predicate. Tests cover translated actions, unmatched terms and timestamp noise.
3. **Test structure:** both source-slice `eval()` calls are removed. The item search predicate and Knowledge Base row formatting/search are tested directly through helpers, with source assertions for handler wiring. Translation expectations use the real language loader instead of fixed French wording.
4. **License:** `logs_filter_logic.php` now has the standard TeamPass GPL header.
5. **JavaScript:** the three new declarations in `showItems()` use `const`/`let`.
6. **Purge accounts:** the selector excludes the OTV, TP, SSH and API account IDs through the parameterized list, while retaining administrators.
7. **Failed-login purge:** the current-login/pre-rename limitation is documented beside the query and in the PR description. The SQLite fixture uses `COLLATE NOCASE`, with web/API case variants and an old-login retention case. Its comment clarifies that this models ASCII case folding, not the full MySQL Unicode collation.

CI has completed successfully for this revision, with GitHub Actions testing the PR merge with develop:

- [Quality](https://github.com/nilsteampassnet/TeamPass/actions/runs/34578042696): **1,803 tests and 60,176 assertions pass on each of PHP 8.2 and 8.3**. PHPStan level 4, all 45 login JavaScript tests, the table-prefix guard and the production-autoloader check also pass.
- [CodeQL](https://github.com/nilsteampassnet/TeamPass/actions/runs/34578042712): both analyses and the result check pass.
- [Scrutinizer](https://scrutinizer-ci.com/g/nilsteampassnet/TeamPass/inspections/06e370d8-43ce-4671-a2b3-79a0857e75c6): successful status on `2f7e9cdfa`, reporting seven updated code elements and passing tests.

The existing workflow deprecation and Composer lockfile warnings remain non-blocking. The PR description and validation checklist have been updated for this commit; manual instance/browser validation remains explicitly pending.

PR #5366, opened 2026-09-08
Fix 5364

guerricv:
## Description

Fix the recurring permissions warning for `storage/logs/teampass_background_tasks.lock` reported in #5364 by correcting runtime-file writers, without weakening the integrity or permissions audit.

- Add shared, DB-free helpers for the background-task lock/trigger and the integrity scan/enqueue locks.
- Attempt POSIX permission repair before writing, using `($currentMode & 0640) | 0600`: preserve `0600`, ensure owner read/write, and do not grant group/other access or change the process-wide umask.
- If opening succeeds but only `chmod` fails, log an actionable warning and continue with the existing access. Invalid/replaced targets and actual open/write failures still fail. The permissions audit continues reporting unresolved unsafe modes.
- Reject symbolic links and non-regular targets. Compare descriptor/path type and `dev/ino` before and after permission repair. These checks detect observed replacements; they do not make path-based `chmod` atomic. Runtime directories must remain protected against untrusted writers.
- Open locks without truncation and update the handler PID only after acquiring its exclusive lock.
- Make signal writes non-blocking with `LOCK_EX | LOCK_NB`, distinguishing contention from I/O failure so a web request does not wait or emit a misleading directory-permission error.
- Probe scan status using an existing read-only handle, without creating files or changing permissions.
- Complete the GPL header and update regression tests and runtime-permission documentation.

Runtime data remains excluded from release checksum comparisons but included in the permissions audit. No exclusion rule or `app/files_reference.txt` change is included.

The task-log writer and the existing unlink-after-unlock lifecycle race are intentionally left for separate follow-ups, as requested in review.

## Related issue

Fixes #5364

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (existing behaviour changes)
- [x] Documentation
- [ ] Translation
- [ ] Refactor / maintenance

## How has this been tested?

Review follow-up commit: `de5a9e5e28cfd74f4231b1935bbcb3b5b9423e3c`. Validation checked on 2026-09-10.

Linux CI:

- [PHP 8.2.33](https://github.com/nilsteampassnet/TeamPass/actions/runs/34442930898/job/102761549655): **OK — 1,732 tests, 59,651 assertions**.
- [PHP 8.3.33](https://github.com/nilsteampassnet/TeamPass/actions/runs/34442930898/job/102761549721): **OK — 1,732 tests, 59,651 assertions**.
[...]

PR #5370, opened 2026-09-10
Protect task logs and keep the background-task lock persistent

guerricv:
## Description

This implements follow-up points 7 and 8 from the review of #5366.

**Prerequisite:** this branch includes #5366 at `de5a9e5e28cfd74f4231b1935bbcb3b5b9423e3c`. Merge that PR first; the changes described below are the additional work. The `Fix_5364` branch was not modified.

### Task journal

- Fix `TaskLogger` path resolution: `LOG_TASKS_FILE` is normally absolute and must not be prefixed with `app/scripts/`. Custom absolute paths, native Windows paths and legacy script-relative paths are preserved.
- Reuse the runtime-file helper with an opt-in strict permission policy for log entries. Preserve `0600`, restrict permissive files to `0640` and do not change the process-wide umask.
- Append under an exclusive lock, seek to the current end only after acquiring it, and preserve existing entries. Recheck file identity and permissions after waiting for the lock.
- A configured destination that cannot be opened, secured or written produces an explicit PHP error-log diagnostic. Do not copy potentially sensitive task arguments to that fallback destination; background processing continues.
- Keep disabled logging and the explicitly configured empty-path PHP error-log destination unchanged.
- Preserve the best-effort chmod policy for locks/signals introduced in #5366. No scanner exclusion or release-manifest change.

### Persistent handler lock

- Isolate the existing acquisition/release logic in the DB-free `app/scripts/backgroundTaskLock.php` class so the real implementation can be tested across processes. The handler delegates only those two operations; scheduling and task execution are unchanged.
- Keep the lock file and its inode after release; never unlink it on success, failure or destruction.
- Acquire non-blockingly and write the PID only after exclusive ownership. Preserve explicit unlocking before closing the descriptor (including compatibility with inherited descriptors); an old PID/file does not indicate a running handler.
- Preserve diagnostics for actual open/write failures and tolerate chmod-only failures on otherwise accessible locks.
- Document deployment coordination: old handlers must finish before new code starts because the old release code can still unlink the file.

## Related issue

Follow-up to #5366, review points 7 and 8; related to #5364.

## Type of change

- [x] Documentation
- [ ] Translation
- [ ] Refactor / maintenance

## How has this been tested?

Local Windows validation with PHP 8.2.33 and 8.3.33:

- PHP syntax checks: passed on both versions.
- Targeted tests: **63 tests, 859 assertions, no failures, errors, warnings or notices**, with **28 platform-dependent skips**, on each version. Executed with `--fail-on-warning --fail-on-notice --fail-on-risky`.
- New regression coverage adds 18 cases: logging enabled/disabled, native absolute and legacy relative paths, explicit error-log fallback, creation/recreation under five umasks, mode repair and preservation of 0600/history/inode, unsafe files still reported by the audit, invalid targets/symlinks, and different chmod-failure policies for journals and coordination locks.
- Concurrent append test: three writers, 240 complete distinct entries.
- Persistent-lock tests use real PHP subprocesses and a request/response protocol with a 15-second process timeout, not timing assumptions. Cases cover a pre-opened contender, a third contender, repeated handoffs, preservation of the active PID, repeated/failed acquisition, normal release, and forced termination followed by restart on the same inode.
- The test fixture reads the PID through the owner's descriptor because Windows mandatory locks prevent outsider reads. The same assertions run on Linux.
- Full configured PHPStan level 4: passed. A temporary copy of `settings.sample.php` was used and removed afterwards; no credentials or production configuration were created.
- Table-prefix guard: passed (81 table names checked). Production autoloader checked: no development-package mappings.
- Full suite on both versions: **1,750 tests, 59,820 assertions, 9 failures, 1 warning, 28 skips**. The Windows full suite is **not green**; these are the platform-sensitive failures also encountered before this follow-up, not failures of the targeted runtime tests.
- `git diff --check`: passed. Release manifest, scope/permission-audit rules, Composer metadata and dependencies unchanged.

**Linux CI for this follow-up is pending.** POSIX permissions and symbolic-link cases cannot be fully validated on this Windows workstation. Chmod denial is simulated in an isolated PHP subprocess; a real multi-account ownership scenario is not exercised locally.

No live-instance database/UI tests were run. Admin, manager, standard/read-only roles and personal-folder settings are unaffected and were not exercised through the UI.

### Suggested lab validation

1. Pause cron/other handler launchers and incoming requests that can start background work. Let old handlers and workers finish, then deploy the complete branch, including the new `app/scripts/backgroundTaskLock.php`, and resume operation.
2. Enable task logging and trigger ordinary background work under the normal web/cron account. Confirm the configured journal is written in its intended location, new lines append, and its mode is `0640` or an existing `0600`.
3. Run a Health permissions scan. The managed journal and locks should no longer raise world-access warnings after successful permission repair; an unrelated unsafe runtime file must remain detectable.
4. After normal completion, confirm that the handler lock remains on disk. Start work again and verify it can acquire that same file and update its PID. Do not infer activity from file presence or the saved PID.
5. Run `php _tools/phpunit.phar --do-not-cache-result --filter BackgroundTaskRuntimeTest` on Linux to exercise the permission and deterministic subprocess scenarios. The forced-termination case kills only a test subprocess, never a live task handler.
6. If checking recreation or log rotation manually, stop all writers first. Never remove a lock while a handler can run.

## Checklist

- [ ] The test suite passes (`php _tools/phpunit.phar` — Linux CI pending; Windows limitations documented above)
- [ ] New `app/sources/*.queries.php` files have a matching `public/sources/` proxy shim — not applicable
- [ ] Changes to `teampassclasses` were applied to **both** copies — not applicable
- [x] `app/vendor/composer/` is in its production form — unchanged

- [x] No schema change

No migration or manifest regeneration is needed for this change. Existing accessible locks and journals are repaired on use when possible. Invalid ownership/path settings still require administrator correction.

The installer and historical upgrade initializers only create missing legacy lock/trigger files; they do not remove the retained runtime lock. They are unchanged by this follow-up.

Deploy with old handlers drained as described above. Keep runtime directories protected against untrusted writers: path-based chmod checks are not atomic and this patch does not claim to make arbitrary external unlink/replacement safe.

## Screenshots

Not applicable: no UI change.

### DIFF
--- a/app/scripts/backgroundTaskLock.php
+++ b/app/scripts/backgroundTaskLock.php
@@ -1,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Teampass - a collaborative passwords manager.
+ * ---
+ * This file is part of the TeamPass project.
+ *
+ * TeamPass is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, version 3 of the License.
+ *
+ * TeamPass is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Certain components of this file may be under different licenses. For
+ * details, see the `licenses` directory or individual file headers.
+ * ---
+ * @file      backgroundTaskLock.php
+ * @author    Nils Laumaillé (nils@teampass.net)
+ * @copyright 2009-2026 Teampass.net
+ * @license   GPL-3.0
+ * @see       https://www.teampass.net
+ */
+
+require_once __DIR__ . '/../sources/runtime_files.functions.php';
+
+/**
+ * Own the handler's advisory lock on a trusted local runtime file.
+ *
+ * File presence and an old PID do not indicate activity: only flock does.
+ * Validate the inode after locking and unlink it before unlocking on release.
+ * A contender holding an orphaned descriptor must never start background work.
+ */
+final class BackgroundTaskLock
+{
+    /** @var resource|null */
+    private $handle = null;
+
+    /** @param string $path Trusted configured local path to the handler lock. */
+    public function __construct(private string $path)
+    {
+    }
+
+    /** Acquire exclusively without waiting or changing another owner's PID. */
+    public function acquire(): bool
+    {
+        if (is_resource($this->handle)) {
+            return true;
+        }
+
+        $handle = tpOpenRuntimeFile($this->path);
+        if ($handle === false) {
+            error_log(
+                'Teampass Background Tasks: cannot open a valid lock file "' . $this->path
+                . '" - check that the web server user can write to this directory.'
+            );
+            return false;
+        }
+        return $this->acquireHandle($handle);
+    }
+
+    /**
+     * Take ownership of an opened descriptor, closing it on any failed attempt.
+     *
+     * @param resource $handle The descriptor opened for this acquisition.
+     */
+    private function acquireHandle($handle): bool
+    {
+        if (@flock($handle, LOCK_EX | LOCK_NB) === false) {
+            fclose($handle);
+            return false;
+        }
+
+        // A finishing handler may have unlinked this inode after our open.
+        $stat = @fstat($handle);
+        if ($stat === false || tpRuntimeFileMatchesPath($this->path, $stat) === false) {
+            fclose($handle);
+            return false;
+        }
+
+        $pid = (string) getmypid();
+        if (@ftruncate($handle, 0) === false || @fwrite($handle, $pid) !== strlen($pid) || @fflush($handle) === false) {
+            fclose($handle);
+            error_log('Teampass Background Tasks: cannot write lock file "' . $this->path . '".');
+            return false;
+        }
+        $this->handle = $handle;
+        return true;
+    }
+
+    /** Remove only our own inode while locked, then release the descriptor. */
+    public function release(): void
+    {
+        if (is_resource($this->handle)) {
+            $stat = @fstat($this->handle);
+            if ($stat !== false && tpRuntimeFileMatchesPath($this->path, $stat)) {
+                unlink($this->path);
+            }
+            // Keep explicit unlocking even if a child inherited the descriptor.
+            // Closing is also the fallback if the explicit unlock fails.
+            flock($this->handle, LOCK_UN);
+            fclose($this->handle);
+        }
+        $this->handle = null;
+    }
+
+    /** Also release when the owner goes out of scope; abrupt exit is handled by the OS. */
+    public function __destruct()
+    {
+        $this->release();
+    }
+}
--- a/app/scripts/background_tasks___handler.php
+++ b/app/scripts/background_tasks___handler.php
@@ -35,6 +35,7 @@ require_once __DIR__.'/../sources/main.functions.php';
 require_once __DIR__ . '/../sources/backup.functions.php';
 require_once __DIR__ . '/../sources/lapr.functions.php';
 require_once __DIR__ . '/taskLogger.php';
+require_once __DIR__ . '/backgroundTaskLock.php';
 
 class BackgroundTasksHandler {
     private array $settings;
@@ -43,7 +44,7 @@ class BackgroundTasksHandler {
     private int $maxExecutionTime;
     private int $batchSize;
     private int $maxTimeBeforeRemoval;
-    private mixed $lockFileHandle = null;
+    private ?BackgroundTaskLock $processLock = null;
     /** @var array<int, array{process: Process, task: array<string, mixed>, resourceKey: ?string}> Running process pool */
     private array $pool = [];
     private string $triggerFile;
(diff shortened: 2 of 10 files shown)


nilsteampassnet:
Thanks @guerricv for the work again.

Your Linux CI was pending, so I ran the branch on Linux (PHP 8.4.22):
- the 38 targeted tests pass, none skipped;
- the full suite is green (1,750 tests);
- PHPStan level 4 reports no errors.

The problems below come from the overlap with `develop`, plus a few adjustments.

### Blocking

**1. The branch conflicts with `develop`, and one file breaks without a conflict marker**

A merge into `develop` conflicts in `background_tasks___handler.php`, `taskLogger.php` and `docs/install/file-permissions.md`. The dangerous file is one git does *not* flag. `app/sources/runtime_files.functions.php` merges cleanly but ends up with two `tpAppendRuntimeFile()` definitions, yours and the one from `fea66e034`:

```
PHP Fatal error:  Cannot redeclare function tpAppendRuntimeFile()
```

`main.functions.php:63` requires that file. Resolving only the three visible conflicts would therefore take down every page and the API.

Please rebase onto `develop`. Change the existing `tpAppendRuntimeFile()` instead of adding a second one. Two tests in `BackgroundTasksProcessLockTest` (`testAcquireChecksDescriptorIdentityAfterLocking` and `testReleaseUnlinksUnderTheLockAndOnlyWhenItOwnsIt`) are tied to the develop implementation; point 8 covers what to do with them.

**2. Please keep develop's lock lifecycle, not a persistent lock file**

Both designs close the race from point 8, as long as every handler runs the same code, so this is a choice. I'd like to keep the design on `develop`:
- the handler unlinks the lock while still holding it;
- it unlinks only when the path still names the inode it owns;
- it checks that identity again right after `flock`.

The reason is that a persistent file turns a temporary problem into a permanent outage:

- The lock file does not exist after an upgrade, because every released version deletes it at the end of each run. It also does not exist on a fresh install before the first run.
- The first run at that moment may not be the web server user. Typically it is an administrator following `docs/misc/troubleshooting.md:31` or `:132`, which run the handler without `sudo -u www-data`. The file is then created owned by `root`, with mode `0640`.
- A persistent lock is never removed, so `www-data` can no longer open it for writing. `acquire()` then fails on every run and the whole background pipeline stops. That includes the sharekey fan-out that new and updated shared items depend on.
- The only trace is one line in the PHP error log. The Health permission scan cannot report it, because it runs in that same background pipeline.

With develop's design, a root run that completes normally deletes its own file, and the next `www-data` run recreates it. A root run killed mid-way leaves the same state with either design, as noted in the #5366 review. The persistent file makes that the outcome of every root run that happens to create the file. What it buys in return is small: a contender that steps aside simply waits for the next cron tick.

The extraction into a DB-free class is worth keeping if you like it, since it is what makes your real-process tests possible. `BackgroundTaskLock` would then implement develop's behaviour:
- identity check after `flock`;
- unlink under the lock, only when the path still matches;
- release does nothing when the lock was never acquired.

The persistent-lock paragraphs and the deployment procedure in the docs then go away.

### Adjustments

**3. Strict log mode: refuse on world access, not on every mode other than 0600/0640**
`tpOpenRuntimeFile(..., true)` (`runtime_files.functions.php:116`) and the re-check in `tpAppendRuntimeFile()` (`:148`) refuse any mode other than `0600`/`0640`. A `0660` log owned by another account is refused, although others cannot read it. It is also exactly the case the documentation tolerates (`file-permissions.md:336`, "a group-writable file belongs to another account").

The goal is that task arguments never land in a world-readable file, and `($mode & 0007) === 0` is enough for that. Please also update that paragraph (`file-permissions.md:330-340`): it currently lists the task log among the files that "continue with existing access".

**4. Report a failed log destination once per process**

The diagnostic is welcome: `develop` currently ignores the return value of `tpAppendRuntimeFile()`. But `taskLogger.php:66` fires on every lost entry, and on two lines when the chmod warning from `tpOpenRuntimeFile()` also fires. The handler has about twenty log calls and runs every minute, and the worker logs too, so a broken destination floods the PHP error log. A static flag in `TaskLogger` that reports it once per process is enough.

**5. Retry instead of dropping the entry when the log is replaced during the wait**

After waiting for the lock, if the path no longer names the same inode (typically logrotate with `create`), the entry is dropped. Reopening once and retrying would keep it.

**6. Reuse `tpResolveRuntimeLogPath()`**

`taskLogger.php:58` re-implements the absolute-path detection inline. It is equivalent to `tpResolveRuntimeLogPath()`, already on `develop` and covered by `TaskLoggerTest`, so please call the helper.

**7. Documentation**

- The deployment procedure (`file-permissions.md:365`) belongs in release notes, not in the permanent guide. It also becomes unnecessary once the persistent lock goes (point 2).
- Please put `LOG_TASKS_FILE`, `enable_tasks_log` and `app/scripts/` in backticks, like the surrounding text.
- Please update develop's existing paragraphs (lines 330–340 and 348–352) instead of adding a parallel description.

**8. Tests**

- `BackgroundTaskRuntimeTest` overlaps `TaskLoggerTest` on `develop` (absolute path, logging disabled, creation mode). Please merge them into one suite.
- Your real-process lock tests are worth more than develop's two tests that only compare `strpos` offsets in the handler source. Adapted to develop's behaviour, they can replace them. The cases would be:
  - a pre-opened contender is rejected by the identity check;
  - the file is removed on a normal release;
  - a file left behind by a killed process can be acquired again.
- Please keep the concurrent-append test (three writers, 240 entries) and the tests proving log contents never leak.

### Scope after the rebase

On top of `develop`, this would leave:
- the once-per-process diagnostic when the task log cannot be written;
- the strict log mode, using the world-access criterion;
- the real-process lock tests, optionally with `BackgroundTaskLock` implementing develop's behaviour.

Thanks again. The test harness in particular is very good work: request/response subprocess protocol, bounded timeouts, and the kill test confined to a test probe.

guerricv:
Thanks @nilsteampassnet. All eight review adjustments are addressed in `2d966dcfe`, now a single commit on top of `develop` (`d0fc7ccb9`).

- **1 / 6:** rebased, resolved the overlaps and extended develop's sole `tpAppendRuntimeFile()` implementation; `TaskLogger` uses `tpResolveRuntimeLogPath()`.
- **2:** adopted develop's lock lifecycle inside the DB-free class: identity check after flock, PID write only after validation, unlink only the owned inode while still locked, and no deletion following failed/repeated release. The persistent-lock behaviour is gone.
- **3 / 4:** strict logging checks `mode & 0007` both on open and after locking. Chmod denial alone no longer refuses an accessible `0660` log. Destination failures are reported once per process, including suppression of duplicate helper diagnostics; lost entries are never copied to the fallback log and later writes still retry the destination.
- **5:** reopen once after a replacement detected at lock handoff; no retry after any bytes have been written.
- **7:** updated the existing documentation, formatting and permission-policy distinction; removed the persistent-lock/deployment paragraphs.
- **8:** consolidated journal tests and retained the 3-writer/240-entry test and no-leak cases. The real lock implementation now has executable orphan-rejection, normal deletion, replacement ownership and SIGKILL/restart coverage; an unlink checkpoint verifies that the native flock is still held.

I also checked the unsafe automatic merge explicitly: PHP 8.2 and 8.3 reproduced the duplicate-declaration fatal before resolution. The final published files match the reviewed local tree.

**CI is green for PHP 8.2 and 8.3: 1,815 tests / 60,632 assertions on each, no skipped tests.** PHPStan level 4, CodeQL, the autoloader/prefix guards and login JavaScript tests also pass:
- [Quality](https://github.com/nilsteampassnet/TeamPass/actions/runs/34598966237)
- [CodeQL](https://github.com/nilsteampassnet/TeamPass/actions/runs/34598966263)
- [Scrutinizer](https://scrutinizer-ci.com/g/nilsteampassnet/TeamPass/inspections/ff790348-4dd0-45f8-a4f2-29a2cfca802e): analysis successful and tests passed.

Release manifests, permission-scan exclusions, dependencies and the production autoloader are unchanged. No live-instance or real multi-account ownership test is claimed; the Linux chmod-denial scenarios use isolated fixtures.

GitHub now reports the PR as conflict-free against develop. The original PR description has been left unchanged; this reply records the revised implementation and completed validation.

PR #5371, opened 2026-09-10
Fix stale folder trees after folder creation

guerricv:
## Description

A manager can successfully create a folder from the Folders management page, yet the new folder remains absent from the Items tree until the user explicitly refreshes the tree. Reloading the page or signing in again can keep displaying the persisted stale cache.

The creation path invalidated the cache, then appended the new folder to only some cached representations and advanced their shared timestamp. The displayed tree (`data`) remained unchanged but could now be considered current. Timestamp-only invalidation also missed changes within the same second. The background cache task could similarly reuse stale tree data, and an API cache write could advance the timestamp without rebuilding the web tree.

This PR makes the next normal folder-tree request rebuild the affected user's cache through the existing permission-aware handlers.

**Changes**

- Introduce `invalidateUserFolderCache()` to clear all three cached representations (`data`, `visible_folders`, and `folders`), reset the build timestamp, and record invalidation. Clearing the payloads prevents same-second changes and partial cache writes from making an old tree usable again.
- Invalidate affected users after folder creation commits and the nested tree is rebuilt. Explicitly include the creator, including personal-folder owners without roles, and include direct folder grants alongside role-based access and administrators.
- Remove the creator's partial cache append, which could not produce a complete tree or reliable folder metadata.
- Change the existing background cache task to invalidate the user's cache. The next request rebuilds it using the regular permission logic instead of deriving lists from stale tree data.
- Store a dropdown-only cache insertion in its requested field, leaving the tree payload empty until the actual tree renderer builds it.
- Make the item-access fallback reuse the current request's permission scope from `core.php` while the cache is unavailable. This preserves personal descendants and read-only metadata, excludes forbidden folders, and refuses to use one user's session for another user ID.

**Behavior and compatibility**

Folder creation permissions, role inheritance, and transaction boundaries are preserved. The correction refreshes cached views of existing permissions; it does not introduce new access grants.

The next normal tree load no longer depends on a WebSocket event or the background worker to discover the creator's new folder. WebSocket events, recipients, and frontend refresh handlers are unchanged. Receiving an existing forced-refresh event can mask the original problem, but enabling WebSocket alone does not guarantee a refresh for the creator.

The shared invalidation helper also serves other folder/cache updates. Unrelated users' cache entries remain untouched by targeted invalidation. A delayed background task may cause one additional cache rebuild on the next request. A completed cache remains reusable until another invalidation.

## Related issue

Related to [discussion #5330](https://github.com/nilsteampassnet/TeamPass/discussions/5330).

## Type of change

- [ ] Documentation
- [ ] Translation
- [ ] Refactor / maintenance

## How has this been tested?

**Automated regression coverage**

Added 26 cases that execute the actual production declarations with isolated database/session doubles and a fixed clock. Coverage includes manager, administrator, and standard-user creation gates; shared, root, and personal folders; personal folders enabled/disabled; personal descendants without roles; all inherited permission types (`W`, `R`, `ND`, `NE`, `NDNE`); multiple roles; direct grants; forbidden and read-only scopes; failed transactions; and unchanged unrelated caches.

Cache-specific cases cover same-second invalidation, missing cache entries, background tasks, API partial writes, dropdown-first requests, and reuse after a complete rebuild.

- **Targeted regression suites:** passed on PHP **8.2.33 and 8.3.33**, with **285 tests and 1,325 assertions** on each version, including all 26 new cases and existing folder/API/security regressions.
- **Full test suites:** ran on both PHP versions with **1,736 tests, 59,436 assertions, 9 failures, 1 warning, and 2 skipped tests** each. The nine failing test names match the pre-change local baseline. These full runs included the first 24 new cases; the final two additional cases passed in the targeted runs above. The full suite is therefore not marked as passing, and Linux CI confirmation remains pending.
- **PHPStan level 4:** passed using the locked PHPStan **2.2.8** standalone PHAR with the repository rules and locally adjusted paths. The temporary settings sample used for analysis was removed afterward.
- **Additional checks:** PHP syntax checks on both versions, the table-prefix guard, the production-autoloader guard, and `git diff --check` passed.

The targeted suites can be rerun with:

```sh
php _tools/phpunit.phar --do-not-cache-result --colors=never --filter 'FolderCacheInvalidationTest|FolderEndpointsRegressionTest|FolderLogicTest|FolderAccessModelTest|PersonalFolderContainmentTest|SearchFolderScopeTest|ApiItemCacheSyncTest'
```

**Manual verification still required**

No live MySQL/MariaDB integration or browser/WebSocket test has been performed. Before merging, verify on a disposable instance:

- [ ] Warm a manager's tree, create a shared subfolder and a root folder from Folders management, then open Items without forcing a tree refresh. Repeat after signing in again and with WebSocket disabled/enabled.
- [ ] Verify creation from Items and visibility for another authorized user, with inheritance enabled/disabled and read-only restrictions preserved.
- [ ] Verify personal descendants for a standard user without roles, personal folders disabled, direct grants, and explicit denials. Other users must not gain access.
- [ ] Run a queued cache task and an API folder-list request between creation and the next web load. Also check API/import creation, rename/move/delete, item counters, and reuse on a second unchanged load.

## Checklist

- [x] PHPStan level 4 passes (`php app/vendor/bin/phpstan analyse --memory-limit=2G`) — equivalent standalone PHAR run described above.
- [ ] The test suite passes (`php _tools/phpunit.phar` — see CONTRIBUTING.md for the one-time setup) — nine pre-existing local failures; Linux CI confirmation pending.
- [ ] Every new public function has a docblock — the new production helper is documented; test-only doubles and PHPUnit methods do not all have individual docblocks.
- [x] No `var_dump()` or `console.log()` left in the code — none introduced by this change.
- [x] New `app/sources/*.queries.php` files have a matching `public/sources/` proxy shim — not applicable; no new query endpoint.
- [x] Changes to `teampassclasses` were applied to **both** copies (`app/includes/libraries/` and `app/vendor/`) — not applicable; neither copy changed.
- [x] `app/vendor/composer/` is in its production form (`git checkout -- app/vendor/composer/`) — verified unchanged.

- [x] No schema change

No installer, dependency, or configuration changes are required. The existing cache columns and background task type are reused. The fix does not perform a global cache purge during upgrade: a cache already stale before deployment may still require the existing tree-refresh button once. Subsequent invalidations use the corrected behavior.

## Screenshots

Not applicable — backend cache and permission-metadata changes; no UI layout changes.


nilsteampassnet:
Thanks @guerricv for this PR and the very detailed write-up. The diagnosis is right, and I confirmed each root cause in the code: `updateUserFolderCache()` advancing the creator's `timestamp` right after the invalidation, the background task and the API writes revalidating an old tree, and the dropdown JSON stored as tree `data` on a cache miss.

On my side, the full suite is green on PHP 8.2, 8.3 and 8.4 (1,738 tests), so the 9 failures you saw locally are specific to your environment. PHPStan level 4 is clean on the four production files and the table-prefix guard passes.

Before merging, I would like the following five points addressed.

### 1. Refresh the permission scope in the fallback, and fix its comment

`buildVisibleFoldersOnTheFly()` now builds the fallback from the session scope. That scope is not resolved "for this request": `items.queries.php` does not load `core.php`, so the session lists date from the user's last full page load (or the last uncached `refresh_visible_folders`). The fallback feeds `getCurrentAccessRights()`, the authorization check for item access. On `develop` it read roles live from the database, so a rights change made by an administrator (for instance removing a role in `users.queries.php`, which deletes the user's `cache_tree` row) applied on the next item request. With this PR it only applies once the session scope is refreshed.

The new fallback is otherwise better aligned with the rest of the authorization model, so I would like to keep it and make it fresh:

- refresh the scope from the database before building the fallback, reusing the block from `refresh_visible_folders` (`items.queries.php` ~L7482-7533: user query, `identifyUserRights()`, root-folder right). Please extract it into a helper used by both places rather than duplicating it;
- run that refresh at most once per request (a static flag), since some handlers call `getCurrentAccessRights()` twice (source and destination of a move);
- update the docblock and the inline comment accordingly;
- add a test: the session still lists a folder the database no longer grants, and the fallback excludes it.

### 2. Drop the similar-roles refresh from `createFolder()`

`invalidateCacheForFolderUsers($newId, [$user_id])` already covers every user who can see the new folder (roles on it, direct grants, administrators, the creator). The `refreshCacheForUsersWithSimilarRoles()` call that follows is therefore redundant, and it is now broader than before. Its tasks wipe the cache instead of refreshing it, and for an administrator `user-roles` holds every role (`identAdmin()`), so `getUsersWithRoles()` returns every non-admin user with a manual role. A single folder creation by an admin ends up wiping the tree of all those users, which contradicts "unrelated users' cache entries remain untouched". The tests do not catch it because `getUsersWithRoles()` is stubbed to return `[8]`.

Please:

- remove the call from `createFolder()`, and the now-unused `refreshCacheForUsersWithSimilarRoles` option from its three callers (`add_folder` in `folders.queries.php`, `FolderModel::createFolder()`, `import.queries.php`);
- keep it in `updateFolder()` (parent change) for now;
- update the two sentinels in `tests/Unit/Api/FolderEndpointsRegressionTest.php` that expect it (`testCreatePassesFullLifecycleOptions` and `testFolderWritesAreTransactional`), for example by asserting instead that `createFolder()` invalidates after the commit and after the tree rebuild.

### 3. Load the fallback folders in one query

The fallback still runs one `SELECT … WHERE id = %i` per folder of the scope, and it is reached often: `adjustFolderItemsCounter()` invalidates every user with access to a folder on each item add or removal. Please load the rows in a single query, for example:

```php
if ($visibleFolderIds === []) {
    return $html;
}
$rows = DB::query(
    'SELECT id, title, parent_id FROM ' . prefixTable('nested_tree') . ' WHERE id IN %li',
    $visibleFolderIds
);
```
(`personal_folder` is no longer needed there, since `perso` now comes from the session.)

### 4. Align the tests with the conventions of `tests/`

The `token_get_all()` + `eval()` harness does run the real code, but it is a new pattern in this repository and a brittle one: the `DB` double matches SQL fragments and throws on any other query, so a harmless refactor of these functions breaks the tests, and redefining `time()` or `error_log()` inside a namespace is easy to miss when reading them. Please follow the conventions already used here:

- move the DB-free decisions into a small logic module required by `main.functions.php`, like `item_revisions_logic.php` or `item_restriction_logic.php` (for example: user-ID normalization for the invalidation, the fallback scope computed from the session arrays, the row inserted by `cacheTreeUserHandler()` on a cache miss), and unit-test it directly;
- guard the wiring with static source sentinels, as `tests/Unit/Api/FolderEndpointsRegressionTest.php` does (commit, then tree rebuild, then invalidation in `createFolder()`; `performVisibleFoldersHtmlUpdate()` (commit, then tree rebuild, then invalidation in `createFolder()`; `performVisibleFoldersHtmlUpdate()` delegating to `invalidateUserFolderCache()`; `users_groups` included in `invalidateCacheForFolderUsers()`);
- then remove `tests/Stubs/folder_cache_runtime.php`.

### 5. Close the remaining gaps in the cache lifecycle
example: user-ID normalization for the invalidation, the fallback scope computed from the session arrays, the row inserted by `cacheTreeUserHandler()` on a cache miss), and unit-test it directly;
- guard the wiring with static source sentinels, as `tests/Unit/Api/FolderEndpointsRegressionTest.php` does (commit, then tree rebuild, then invalidation in `createFolder()`; `performVisibleFoldersHtmlUpdate()` delegating to `invalidateUserFolderCache()`; `users_groups` included in `invalidateCacheForFolderUsers()`);
- then remove `tests/Stubs/folder_cache_runtime.php`.

### 5. Close the remaining gaps in the cache lifecycle

**a) Update and copy paths.** `FolderManager::updateFolder()` (it already receives `user_id` in `$params` from the API), and `update_folder` and `copy_folder` in `folders.queries.php`, still call `invalidateCacheForFolderUsers()` without the acting user. An owner without roles who renames or moves a personal subfolder therefore keeps the stale tree you fixed for creation. Please pass the acting user there too.

**b) `user_build_cache_tree` producers.** The task now only invalidates. The login producer (`identify.php`, when the row is missing or `visible_folders` is empty) and the one queued at the end of user creation (`UserHandlerTrait`) no longer build anything: at best they are a no-op, at worst they wipe a tree built in between. Please remove them. The import producer (end of `import.queries.php`) can become a direct `invalidateUserFolderCache([$userId])` after the tree rebuild. Please keep the worker case so that tasks already queued at upgrade time still complete.

**c) Build/invalidation race.** A tree build that starts before an invalidation and is written after it still wins, because the write resets `invalidated_at` to 0 unconditionally. Please capture the build start before building (`tree.php`), store it as `timestamp`, and persist only when no invalidation happened since (`… WHERE user_id = %i AND IFNULL(invalidated_at, 0) < %i`). A refused write simply leaves the cache empty and the next request rebuilds it. Same-second invalidations are then refused too, which costs at most one extra rebuild.

Once these are in, I will run the manual checks listed in your PR on a test instance. Happy to discuss any of these if you see a better approach.
Thanks again for the careful work on this one!

guerricv:
Hi Nils,

Thanks for checking the implementation in detail. Your five points were valid, including the incorrect assumption in my comment about `core.php` being loaded for each item request.

The follow-up changes address them as follows:

1. **Refresh permission scope from the database.** The fallback and uncached dropdown now share `refreshUserFolderPermissionScope()`, which reuses `identifyUserRights()` and runs at most once per request. It reloads manual/AD roles, forbidden folders, personal-folder availability and the root-folder creation right. The role arrays consumed by item restrictions and role-based authorization are refreshed as well.
2. **Remove the broad refresh after creation.** `createFolder()` now relies on targeted invalidation, and the obsolete option is removed from the web, API and import callers. The parent-change path in `updateFolder()` is retained as requested. The API lifecycle sentinels now assert commit, tree rebuild, then invalidation.
3. **Batch the fallback query.** Folder metadata is loaded with one `WHERE id IN %li` query, with an early return for an empty scope. Personal and read-only metadata still come from the refreshed permission scope.
4. **Use the existing test conventions.** DB-free decisions are in `folder_cache_logic.php`, included by production and tested directly. Static sentinels cover the lifecycle and authorization/cache wiring. The `token_get_all()`/`eval()` fixture and its namespace overrides have been removed.
5. **Complete the cache lifecycle.** Update/copy paths explicitly invalidate the acting user, including personal-folder owners without roles. Login and user-creation cache tasks are removed; import invalidates directly after rebuilding. The legacy worker still consumes already queued tasks. Cache writers use the build start time and conditional updates.

Two implementation details needed a little more coverage than the initial outline:

- **Authorization before the fallback:** `getCurrentAccessRights()` can return from direct-grant or read-only checks before reaching the fallback. It now refreshes the scope and checks folder membership before those shortcuts and item-restriction checks. This also prevents a stale role list from surviving a role removal. The refresh is performed once per authorization request, including when the visible-folder cache is warm; subsequent checks in that request reuse it.
- **Invalidation across web/API writers and missing rows:** API writes also used to clear `invalidated_at`. All tree, dropdown and API writes now share the same guard and retain the invalidation marker. A build captures its cache row ID and marker before reading permissions/folders; on a first load, an empty row is initialized before those reads. The write must still target that row, observe the same marker, and be newer than its invalidation; it cannot overwrite a cache with a later build timestamp. A row deleted during a build is therefore not recreated with the old payload. Same-second invalidations reject the write and leave a later request to rebuild. Since the existing schema permits duplicate user rows, readers consistently select the first row; no schema migration is introduced.

During the lifecycle review, I also aligned folder deletion with commit → tree rebuild → invalidation, preserving the actor and direct-grant recipients collected before deletion. Publishing invalidation inside the transaction could otherwise precede the visibility of the updated tree.

The changes retain the existing folder-creation/inheritance rules and WebSocket events. They do not require a schema, dependency or installer change. The extra rights refresh is intentional for authorization correctness; the fallback no longer adds one metadata query per accessible folder.

Local validation on PHP **8.2.33 and 8.3.33**:

- Full runs: **1,736 tests, 59,531 assertions**, with the same **9 pre-existing Windows failures**, 1 warning and 2 skipped tests on each version. I compared the failing test names against the baseline: no new failures.
- Within those runs, the folder/cache/API/security regression subset is fully green: **283 tests, 1,416 assertions**, including the replacement logic tests and lifecycle sentinels.
- **Full PHPStan level 4 passed**, using the locked 2.2.8 standalone PHAR and the repository rules.
- PHP syntax checks pass on both versions; the table-prefix and production-autoloader checks pass.

The previous green GitHub CI applies to the earlier commit; this follow-up still needs its own CI run and the manual checks on your test instance, especially the concurrent writer/deletion cases against MySQL/MariaDB.

PR #5378, opened 2026-09-13
Move renewal to users and add per-item password expiration

guerricv:
## Description

Renewal was exposed as an administrative utility even though it operates on items that administrators should not access. This PR moves it to the user area and adds optional password renewal periods on individual items, with consistent deadlines across item views, Security Posture and governance reports.

Target: `nilsteampassnet/TeamPass:develop`  
Source: `guerricv/TeamPass:checking_utilities.renewal`  
The branch was aligned with upstream `develop` at `aa8d66bd7` before implementation. Latest commit: `0189bee80`.

- **User access and permissions:** Renewal is available to non-admin users, managers, HR managers and read-only users. Administrators are denied both the page and its data endpoint, including accounts with combined privilege flags. Rows, search results, counts, pagination and folder paths respect current database permissions, item restrictions and personal-folder ownership. Existing administrative compliance reports retain their details; this PR does not remove item names from those reports.
- **Optional item policies:** Editors can set an individual renewal interval of 1–36500 days on personal or shared items, independently of the global folder-expiration switch. The shortest active item/folder interval wins. Disabling the individual setting preserves any active folder rule. The menu remains available when folder expiration is off.
- **Password age and API:** Deadlines use the last password change or creation. Policy edits and moves preserve password age; copies inherit the period and start a new age. Policy changes have a separate audit entry. API reads, creation and updates support `renewal_period`; omitted update values preserve the policy, and existing idempotency intents remain compatible. Expiration does not delete data, although existing consultation restrictions can apply.
- **Visible policies and deadlines:** Folder banners and item forms explain the applicable policy, including empty folders. Creation/copy estimates and move previews show the destination deadline. Renewal initially lists all known deadlines, expired items first; an optional inclusive date cutoff can be cleared. Item cards show a dated status badge. The main list uses security-style icons to the left of the item name: an orange hourglass within 14 days, or a red crossed calendar after expiration, with escaped tooltips. Unknown password age remains explicit.
- **Posture and governance:** Ordinary overdue and missing-expiration indicators use the effective policy. Shared-folder coverage distinguishes individual policies, covered items, effective overdue items and folder SLA breaches. Personal folders and their descendants are excluded from governance renewal reports. The 14-day look-ahead is shared with the badges.
- **LAPR exclusion:** While LAPR is enabled, ordinary expiration excludes all linked items: managed account passwords and endpoint connection credentials. Paused/error accounts and disabled/error endpoints remain excluded, independently of the scheduler and policy assignment. This applies to Renewal, previews, badges, search/favourites and expiration-related posture/governance counts, including “no expiration.” Renewal coverage totals exclude these items; other security checks and existing rotation evidence retain their scope. The form hides ordinary renewal controls, and web/API updates reject policy changes while allowing unchanged values. Dormant periods are preserved: removing the last link or disabling LAPR restores ordinary renewal using the existing password age. Copies do not inherit LAPR relationships.
- **Translations and documentation:** All 35 added/revised renewal messages are supplied in all 25 language catalogs. Tests check catalog loading, UTF-8, unique keys and substitution tokens. Feature/API documentation and JavaScript test guidance are updated.

**Review is required before merging. This is a substantial PR (69 changed files, including 25 translation catalogs). Additional tests may be needed to fully validate the behavior, particularly permissions, install/upgrade, LAPR integration and report consistency.**

## Related issue

No linked issue.

## Type of change

- [x] Breaking change (existing behaviour changes)
- [x] Documentation
- [x] Translation
- [ ] Refactor / maintenance

Behavior changes include moving Renewal out of the administrator area, displaying all deadlines by default and excluding LAPR-linked items from ordinary expiration.

## How has this been tested?

- PHP 8.2.33 and 8.3.33: **181 tests / 5202 assertions passed** in the final targeted runs. Coverage includes renewal authorization and SQL, policy precedence, password age, migration replay, API policy guards, translations, posture and search. The SQL-backed permission fixture exercises standard, manager, HR, read-only and admin roles, combined flags, revoked grants and personal-folder settings. The LAPR matrix executes the production relation helper and renewal/report queries across link statuses and module on/off.
- Additional password-health consistency checks on PHP 8.2: **8 tests / 87 assertions passed**.
- All JavaScript suites: **63 tests passed**, including 12 Renewal tests for initialization, date cutoff/clear, preview ordering, form controls, badge states/escaping and LAPR notices. Items/Search/Reports templates also passed syntax checks using inert PHP substitutions.
- Full **PHPStan level 4 passed** with the repository configuration. PHP syntax checks, table-prefix guard, diff whitespace checks, dual-copy consistency and production Composer autoloader verification passed.

These checks use controlled HTTP/DOM boundaries and SQLite for database-backed unit tests. **No live TeamPass browser integration, full fresh installation or MySQL/MariaDB upgrade was run.** SQLite does not reproduce every MySQL SQL mode, collation or ALTER TABLE behavior.

The complete PHP suite is **not claimed to pass**. Earlier full runs identified existing OpenSSL setup and LogsFiltering errors plus unrelated source/environment-sensitive failures; the failing names were compared with the upstream baseline. A broad selection during the final work also encountered the existing LicenceSignatureTest OpenSSL setup error. The final targeted runs above are green.

Before merging, please review and supplement testing on a configured instance: fresh install and upgrade/replay; MySQL/MariaDB with `ONLY_FULL_GROUP_BY`; permissions and revoked access; personal/shared creation, edits, password changes, copies and moves; competing periods with folder expiration on/off; date filters and list/card tooltips; managed and credential-only LAPR links, paused states, unlinking and module off; and agreement between Renewal, Posture and governance outputs. Native-speaker review of translations would also help.

## Checklist

The debug-output checkbox applies to this PR's additions. There is no new query endpoint requiring a proxy. PHPStan was run through the standalone PHAR with `phpstan.neon`; the production autoloader was left unchanged. The full-suite checkbox remains open for the reasons above.

- [ ] No schema change

Adds `items.renewal_period INT UNSIGNED NOT NULL DEFAULT 0`. Fresh-install DDL is updated; `public/install/upgrade_run_3.2.2.5.php` is called from the 3.2.2 upgrade chain before saving the upgrade timestamp. Version metadata is set to 3.2.2.5 and the schema floor is raised so Docker's patch replay uses the same migration. Please coordinate the patch number with the release plan.

Existing items receive no individual policy automatically. Migration tests preserve existing rows and configured values on replay, including a non-default table prefix. Run the normal database upgrade before using the new code. The compound schema checkbox remains open because a full fresh installation has not been tested. No dependency or production Composer autoloader changes are included.

## Screenshots

No screenshots attached. Visual validation on a configured instance remains part of the requested review.


nilsteampassnet:
Thanks a lot for this PR @guerricv.
Two changes are real improvements: `renewalEffectiveDays()` / `renewalPeriodSql()` replace the duplicated expiry expressions, and `expired_item` is now computed by the server instead of trusted from the client.

A few points before merging:

### 1. The Docker upgrade stops at `upgrade_run_3.2.2.5.php` (blocking)

`upgrade_chain()` in `docker/docker-entrypoint.sh` runs every `upgrade_run_*.php` file whose name matches `*.*.*`. `3.2.2.5` matches, and `version_to_number()` gives it the same number as `3.2.2`. So when a container upgrades a 3.2.1 database to the 3.2.2 image, it runs `3.2.2`, then `3.2.2.5` as a separate step:

1. that step does nothing, because the file only declares a function;
2. `app/scripts/write-db-version.php 3.2.2.5` then exits with code 1, because it only accepts `X.Y.Z` (`^[0-9]+(\.[0-9]+){1,2}$`);
3. the container logs "Upgrade stopped at 3.2.2.5" and keeps the install directory.

Suggested fix: move the `addColumnIfNotExist(prefixTable('items'), 'renewal_period', 'INT UNSIGNED NOT NULL DEFAULT 0')` call into `upgrade_run_3.2.2.php` itself (that's how `items.revision_changed_at` was added in 3.2.2.1), and delete the separate file. `RenewalMigrationTest` currently reads `upgrade_run_3.2.2.5.php`, so it will need to change too.

### 2. Expired items and viewing access: a proposal

Today, when an item is expired, `show_details_item` returns `show_detail_option = 2` to users without edit rights. `showDetailsStep2()` is then skipped, so the password and details never load. Before this PR, that could only happen if an administrator had enabled `activate_expiration` and the folder had a renewal period.
With individual policies, any editor can set a 1-day period on a shared item, and the next day the other users can no longer open it. No admin setting is involved.

An admin setting to switch individual policies on or off would defeat the purpose of this PR, since these policies are designed to work regardless of the folder expiration setting. So my proposal is to **stop  blocking viewing entirely**. Expiration would stay a visible signal that asks for a password change (list icon, card badge, Renewal page, Security Posture and governance reports), but it would never stop an authorized user from reading the item.

This would also remove a path that is already broken: the "expired" layout targets `#item_details_expired_full`, which no longer exists in `items.php`, so today the item card stays half-loaded. The `renewal_notice_explanation` string ("Consultation may be restricted…") would need updating to match.

What do you think?

### 3. A renewal banner on every folder

`tpRenewal.update('#folder-renewal-notice', groupe_id)` runs every time a folder is listed. When no policy applies, the banner still shows "No password renewal deadline in this folder.", even on instances that never use expiration. Each call also adds a request that runs `securityPostureAuthorizedFolderIds()` (about 5 queries, including a self-join on `nested_tree`). Could the banner only appear when a policy is active (`days > 0`)? It would be even better to skip the request when nothing can apply, for example by returning the folder period in the existing list response.

### 4. Smaller points

- `app/sources/reports.queries.php`: `getPersonalFolderIdsWithDescendants()` and the `NOT IN (...)` fragment are computed at the top of the file, before `PerformChecks`, for every report type. Could they move into the three cases that use them (`report_posture_summary`, `report_rotation_overdue`, `report_rotation_sla`), with `%li` binding instead of a literal list?
- `renewalEligibleItemSql()`: when LAPR is enabled, it adds a correlated `NOT EXISTS` on `lapr_endpoints.ssh_credential_source`. That column has no index, and the subquery runs 2–3 times per row in the dashboard, report and search expressions. The tables are small, so the impact is limited. Still, an index on that column (in `upgrade_run_3.2.2.php` and `run.step5.php`) or a one-time list of excluded item ids would be cleaner.
- `renewalApplicablePeriodSql($settings, $item, $folder)` accepts custom column references, but always calls `renewalEligibleItemSql($settings)` with the default `i` alias. Every current caller uses `i`, so nothing is broken today, but passing the item alias through would avoid a future trap.
- `find.queries.php`: `SELECT c.*, … AS renewal_period, … AS timestamp` returns duplicate column names and relies on the last one winning. Explicit aliases (e.g. `effective_renewal_period`, `renewal_base_date`) would be more robust.
- `app/api/openapi.json` doesn't document `renewal_period` on item get/create/update (`docs/api/api-basic.md` does), nor the new `409` for LAPR-linked items.
- Minor: the move confirmation uses `window.confirm()` instead of the usual modal pattern, and `require_once __DIR__ . '/renewal_logic.php'` sits above the file header in `search.functions.php`.

Happy to discuss any of these.

guerricv:
Thanks for the detailed review, Nils!

We agreed with your proposals and pushed the corrections in df0a5b8fb, after aligning the branch with the latest develop and resolving the conflicts.

The changes address all the points you raised:
- Docker upgrade: moved the renewal_period migration directly into upgrade_run_3.2.2.php and removed upgrade_run_3.2.2.5.php. The migration remains idempotent and preserves existing policies. Regression coverage now includes Docker’s migration discovery and step runner.
- Viewing expired items: we adopted your proposal. Expiration is now advisory for both individual and folder policies: authorized users, including read-only users, can continue reading expired items. Editing permissions remain enforced. The list icons, card badges, Renewal page, Security Posture and governance indicators are preserved. The obsolete expired-item layout has been removed, and the explanation has been updated in all 25 languages and in the documentation.
- Folder banner: it now appears only when an active folder policy applies. It reuses the folder data already loaded by the item list, avoiding the separate preview request. Empty folders are covered, and stale responses cannot restore an outdated banner. Individual item policies remain independent of folder expiration.
- Reports: personal-folder exclusions are now computed only inside the three relevant report cases, after authorization, with %li binding and explicit handling of an empty exclusion list.
- LAPR lookups: added an index on ssh_credential_source for fresh installations and existing databases. It is deliberately non-unique because multiple endpoints can share a connection credential. All LAPR-linked item exclusions remain in place.
- SQL helpers and API: the item alias is now passed through to the LAPR predicate; Find uses distinct aliases for calculated renewal values. OpenAPI now documents renewal_period, omission/reset semantics and the LAPR-related 409 response.
- UI and cleanup: move confirmations now use an asynchronous Bootstrap modal, with cancellation and dismissal handling. The renewal include was also moved below the Search file header.

All nine CI checks pass on the pushed commit, including the PHP 8.2/8.3 suites, JavaScript tests, PHPStan and CodeQL.
A further review and functional testing on a staging instance would still be valuable, particularly for upgrade/replay scenarios and the expired-item and move workflows. No live Docker/MySQL upgrade was performed locally.

😀😀

